PR #5, opened 2015-01-29
Update for nightly 92ff8ea52 2015-01-27

hugoduncan:
Note: I have only checked this compiles, as I am running on OS X.


diwic:
Cool, thanks for helping out!
I merged it just to have it compiling, but when looking it through I wonder if the result value from `write!` is discarded, where it instead should have been returned?

PR #6, opened 2015-02-03
Add support for Struct and ObjectPath types

hugoduncan:
Adds MessageItem variants for Struct and ObjectPath.


hugoduncan:
Happy to squash if this looks ok.

diwic:
Thanks, merged now! I'm not too picky about squashing at this point, just happy for contributions. :-)

Btw, are Struct and ObjectPath types also sent over the protocol, if so you might want to add them to the message_types test as well.

PR #7, opened 2015-02-03
Update for rustc f1398d273 2015-01-31

hugoduncan:
rustc 1.0.0-nightly (f1398d273 2015-01-31 20:21:02 +0000)


diwic:
Thanks for looking after it!

PR #8, opened 2015-02-04
Add derive(Debug) on Connection

hugoduncan:
This is to allow users to add Debug on structs that contain a Connection.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -470,11 +470,14 @@ impl<'a> Iterator for ConnectionItems<'a> {
 /* Since we register callbacks with userdata pointers,
    we need to make sure the connection pointer does not move around.
    Hence this extra indirection. */
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


diwic:
If it helps you I guess I could merge it, but I wonder if it's good or bad to add it just to aid structs that contain a Connection - OTOH, if the Connection/IConnection suddenly wanted another struct member that does not derive Debug then we would have to remove derive(Debug) again, or implement a custom Debug.

Anyhow, maybe it makes more sense to actually implement a useful Debug that writes e g the unique name and what bus it is connected to (e g System or Session bus). What do you think?

diwic:
> Anyhow, maybe it makes more sense to actually implement a useful Debug that writes e g the unique name 

I just did this instead, so I think this pull request can be dropped. Thanks for the interest though :-)

hugoduncan:
Happy to see your suggestion implemented - you beat me to it :)

PR #11, opened 2015-04-01
Use into_cow feature to fix compile errors in latest Rust

kylc:
Library currently fails to compile due to feature-gated `into_cow`. Adding the feature fixes the errors.


diwic:
Thanks!

PR #12, opened 2015-04-29
#[deprecated] is now blocked outside stdlib

Ryanl92:
https://github.com/rust-lang/rust/issues/22830 has caused this.
I've changed the #[deprecated] to doc comments to let it compile. Not sure the best way to tell library users about deprecation now though.


diwic:
Thanks for looking after it! I should probably release a new version on crates.io too. Or maybe I should just remove the deprecated ones, after all, they are deprecated :-)

PR #14, opened 2015-05-18
Make MessageItem::new_array() return a Result

manuels:
As rust main intention is to be secure, the behavior of the code should be reflected by the code and instead of the documentation.
This patch makes `MessageItem::new_array()` return a `Result` instead of panicking i.e. when the vector is emtpy, so you must explicitly call `MessageItem::new_array().unwrap()` in your code which makes potential side effects obvious.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -3,6 +3,12 @@ use std::{fmt, mem, ptr};
 use super::{ffi, Error, MessageType, TypeSig, libc, to_c_str, c_str_to_slice, init_dbus};
 use std::os::unix::io::{RawFd, AsRawFd};
 
+#[derive(Debug,Copy,Clone)]
+pub enum ArrayError {
+    EmptyArray,
+    DifferentElementTypes,
+}
+
 fn new_dbus_message_iter() -> ffi::DBusMessageIter {
     ffi::DBusMessageIter {
         dummy1: ptr::null_mut(),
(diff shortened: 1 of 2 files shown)


src/message.rs
```diff
@@ -3,6 +3,12 @@ use std::{fmt, mem, ptr};
 use super::{ffi, Error, MessageType, TypeSig, libc, to_c_str, c_str_to_slice, init_dbus};
 use std::os::unix::io::{RawFd, AsRawFd};
 
+#[derive(Debug)]
+pub enum ArrayError {
```
diwic: ArrayError should derive Copy and Clone, and probably needs to be reexported as well?

src/message.rs
```diff
@@ -204,10 +210,19 @@ impl MessageItem {
 
     /// Creates an MessageItem::Array from a list of MessageItems.
     /// Note: Will panic if the vec is empty or if there are different types in the array
-    pub fn new_array(v: Vec<MessageItem>) -> MessageItem {
```
diwic: The "will panic" comment needs to change accordingly.

src/message.rs
manuels: Fixed the docs and the issue with Clone and Copy.

> probably needs to be reexported as well

I'm not sure what you mean

diwic:
There is a line in src/lib.rs:

    pub use message::{Message, MessageItem, OwnedFd};

The question is if this line also needs to include `ArrayError` so that clients can match on what type of error occurred?

manuels:
Just in case: I just added it to the exports.

diwic:
Thanks, merged. That said, I've been thinking today that maybe it would be better to actually make a new array struct, that always upholds the invariant that all items are the same type. But that'll be something for another day.

PR #18, opened 2015-08-20
Compliance to RFC1214

hoodie:
Hi, the current 1.4 nightly introduces [RFC1214](https://github.com/rust-lang/rfcs/blob/master/text/1214-projections-lifetimes-and-wf.md#impact-on-cratesio) which causes `trait MessageItem` to produce a warning. This warning will become a hard error in the next release. Here is the fix, tests still run.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -402,7 +402,7 @@ impl From<(MessageItem, MessageItem)> for MessageItem {
 }
 
 /// Helper trait for `MessageItem::inner()`
-pub trait FromMessageItem<'a> {
+pub trait FromMessageItem<'a> :Sized {
     fn from(i: &'a MessageItem) -> Result<Self, ()>;
 }
 


.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+*.swp
```
diwic: What does this gitignore addition mean, is it related to RFC1214?

src/message.rs
```diff
@@ -402,7 +402,7 @@ impl From<(MessageItem, MessageItem)> for MessageItem {
 }
 
 /// Helper trait for `MessageItem::inner()`
-pub trait FromMessageItem<'a> {
+pub trait FromMessageItem<'a> :Sized {
```
diwic: This part looks good, thanks for helping out!

.gitignore
hoodie: Ups, sorry, that slipped in. It ignores vim swapfiles, my bad :)

PR #22, opened 2015-09-11
objpath: implement support for annotations

mathstuf:
Properties, methods, and signals may all contain properties according to
the spec. In fact, there are standard properties used to indicate
whether symbols are deprecated, whether a method has no reply, or when a
property signal is emitted for changed properties.

### DIFF
--- a/src/objpath.rs
+++ b/src/objpath.rs
@@ -16,15 +16,28 @@ impl<'a> Argument<'a> {
     }
 }
 
+struct Annotation {
+    name: String,
+    value: String,
+}
+
+struct ISignal<'a> {
+    args: Vec<Argument<'a>>,
+    anns: Vec<Annotation>,
+}
+
 /// Declares that an Interface can send this signal
 pub struct Signal<'a> {
     name: String,
-    args: Vec<Argument<'a>>,
+    i: ISignal<'a>,
 }
 
 impl<'a> Signal<'a> {
     pub fn new<N: ToString>(name: N, args: Vec<Argument<'a>>) -> Signal<'a> {
-        Signal { name: name.to_string(), args: args }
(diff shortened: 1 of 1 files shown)


mathstuf:
One question is whether `::new` for each struct should have an annotation argument (bumping the version number).

mathstuf:
Also makes me wonder if property read/write methods should take a `Connection` to emit the `org.freedesktop.DBus.Properties.PropertiesChanged` signal.

diwic:
This looks really good, so I merged it, thanks!

 > One question is whether ::new for each struct should have an annotation argument (bumping the version number).

Maybe a builder approach would be good for this scenario? (I e, `annotate(self) -> Self` or possibly `annotate(&mut self) -> &mut Self`) 
You can add that as a separate pull request if you like the idea.

 > Also makes me wonder if property read/write methods should take a Connection to emit the org.freedesktop.DBus.Properties.PropertiesChanged signal.

Yeah, I've been talking about the `PropertiesChanged` signal in some other issues. But I doubt this is the right approach? I believe the most common case for wanting to send a `PropertiesChanged` signal is when a property is changed locally and you want to inform the other side. It might very well be read-only to the other side. In that case, no method is called.

mathstuf:
Ah, yes, a builder approach would be good for this. In fact, that might be good for arguments and such as well…would get rid of lots of unnamed `vec![]` arguments scattered about.

diwic:
I've started sketching on a builder approach now.

diwic:
Ok, so I've now started rewriting the `obj` module to fix this and many other things. It uses a builder approach for annotations and a lot of other things. It's still WIP and does not fully work yet, but you can have a look at the new `mdisp` module if you're curious.

PR #23, opened 2015-09-11
Add peer interface

mathstuf:


diwic:
Hi! 

The code itself looks good, but I have two questions.

1) Should this really be implemented on every object path, based on whether they're introspectable or not? I looked around the existing D-Bus services on my machines, and the majority of services did not implement Peer.

2) According to the  [specification](http://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer): "It does not matter which object path a GetMachineId is sent to. The reference implementation handles this method automatically." Does that affect us? Do we even need an implementation?

mathstuf:
It really should show up in the introspection XML. And as such, adding it to the introspection block would be probably be better.

diwic:
Hi, going through old stuff that's lingering. Still unsure what to do with this one, but I think I would accept this patch only if rewritten to 1) support the `tree` module instead of the now deprecated `obj` module and 2) opt-in through a separate `peer()` call.

PR #25, opened 2015-10-18
Make import Rust 1.3 compatible

manuels:
The current code does not compile on my machine (with Rust v1.3).
Could it be that you are using Rust v1.4?
This pull-request fixed a syntax error in an import statement.


diwic:
I use a-few-weeks-old nightly, I think. Thought I'd be safe if I didn't use any `feature`s. Apparently not so.

Thanks!

manuels:
Thanks, diwic!
A version bump would be nice ;)

diwic:
Done now, enjoy!

manuels:
Thanks!

PR #29, opened 2016-01-12
Unregister already-registered paths if error in set_registered()

agrover:
Fix a FIXME, keep track of registered paths so if one fails to register
we can leave things not in a half-registered state.


diwic:
Thanks!

PR #30, opened 2016-01-12
Fix paste error in comment

agrover:
Pretty sure from context add_p adds a property.


diwic:
Thanks!

PR #31, opened 2016-01-13
Make MethodFn, MethodFnMut, and MethodSync public

agrover:
Tree is parameterized by MCall, so if we want to return a tree we need to
make public the structs that implement MCall.

Also make MCall itself pub. This silences a private_in_public warning
that nightly started emitting.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,6 +78,7 @@ pub mod tree {
     pub use methoddisp::{Factory, Tree, TreeServer, ObjectPath, Interface, Signal};
     pub use methoddisp::{Property, EmitsChangedSignal, Access};
     pub use methoddisp::{Method, MethodErr, MethodResult, Argument};
+    pub use methoddisp::{MethodFn, MethodFnMut, MethodSync};
 }
 
 static INITDBUS: std::sync::Once = std::sync::ONCE_INIT;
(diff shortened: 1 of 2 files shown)


agrover:
Please see [this](https://github.com/agrover/froyo/blob/master/src/dbus_api.rs) for my usage that required these structs to be made pub, in case I'm doing something less than optimally...

agrover:
well I don't know, but here's a somewhat reduced test case:
```rust
#[test]
fn prop_lifetime_simple3() {
    {
        let f = Factory::new_fnmut();
        {

            let mut i = f.interface("com.example.dbus.rs");
            {
                let mut x = 5u32;
                let prop = f.property("changes", 0i32);
                {
                    let prop2 = f.property("setme", 0u8);
                    let prop2 = prop2.on_set(|_,_,_| {
                        x = 9;
                        Ok(vec!())
                    });
                    let _setme = i.add_p_ref(prop2);
                }
            }
        }
    }
}
```

If you move `let mut x` line above `let mut i` then the error goes away, or if the add_p_ref line is commented out.

Ok how about this: after the on_set() then prop2's lifetime depends on values its closure references. When you call add_p_ref then if prop2's closed-over references didn't live as long as `i` then... then... ugg I'm so confused, I don't know.

diwic:
Ok, so I realized that even if properties don't work the way I want, normal methods can still benefit from borrowing. So the `<'a>`'s will remain. And thus I can make them public the way they currently are.

I also renamed Mcall to MethodType, but I doubt that matters since it's still not externally public, but apparently that's not needed to silence the warning (!).

PR #34, opened 2016-01-28
Enable missing_docs warning and add missing docs

agrover:
I got bored this afternoon and turned on the missing_docs warning for
everything except ffi. I've added minimal doc strings to satisfy the
doc check, although of course more complete descriptions would be great
to add at some point, too.

I also took the liberty of re-filling the line breaks of some existing
long docstrings.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,8 @@
 //!
 //! See the examples directory for some examples to get you started.
 
+#![warn(missing_docs)]
+
 extern crate libc;
 
 pub use ffi::DBusBusType as BusType;
@@ -31,6 +33,7 @@ use std::cell::{Cell, RefCell};
 use std::mem;
 use std::os::unix::io::RawFd;
 
+#[allow(missing_docs)]
 mod ffi;
 mod message;
 mod prop;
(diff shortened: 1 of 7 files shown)


diwic:
Cool, thanks!

I guess it's a matter of taste if you like having comments like these or if they're just nuisance:

    /// Register an object path.
    pub fn register_object_path

I don't mind either way, and there are certainly a lot of more useful comments too. So, merged, although I reserve the right to remove the missing-docs warning again if I add something new that I don't care to document :-)

agrover:
Yup, cool

diwic:
> I reserve the right to remove the missing-docs warning again if I add something new that I don't care to document :-)

Which happened...now. :-) Recently I've been a bit annoyed by the fact that MessageItem feels like an unnecessary detour, add the recently discovered fact that dict entries must be inside an array, and the fact that nothing stops you from putting a messageitem of the wrong type into an MessageItem::Array...so I started writing something new. Have a look at the new `arg` module (in msgarg.rs) if you're curious! :-)

agrover:
Will do

PR #36, opened 2016-02-14
Use pkg-config

myfreeweb:
Hi! Please use `pkg-config` to link to native libraries. This is especially important on systems like FreeBSD, where the default linker `/usr/bin/ld` doesn't include `/usr/local/lib` in its default search path.

Also, this pull request fixes this:

```
src/msgarg.rs:545:31: 545:39 error: no method named `as_ref` found for type `&std::ffi::c_str::CStr` in the current scope
src/msgarg.rs:545         let p = sig.map(|s| s.as_ref().as_ptr()).unwrap_or(ptr::null());
                                                ^~~~~~~~
src/msgarg.rs:545:31: 545:39 note: the method `as_ref` exists but the following trait bounds were not satisfied: `std::ffi::c_str::CStr : core::convert::AsRef<_>`
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,8 +10,13 @@ documentation = "http://diwic.github.io/dbus-rs-docs/dbus"
 keywords = ["D-Bus", "DBus"]
 license = "Apache-2.0/MIT"
 
+build = "build.rs"
+
 [dependencies]
 libc = "0.2"
+
+[build-dependencies]
+pkg-config = "0.3"
 
 [dev-dependencies]
 tempdir = "0.3"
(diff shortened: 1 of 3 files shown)


diwic:
Thanks!

That as_ref stuff is weird - it compiles fine here (rustc 1.8.0-nightly (3c9442fc5 2016-02-04)) - but anyhow it works fine without it, too.

myfreeweb:
Ah, you're on nightly. I'm on stable. Maybe they've added the trait impl since stable…

diwic:
So if my code doesn't compile on Rust stable...that seems like a good reason to release v0.3.2. Done so now!

PR #41, opened 2016-04-21
Add Tree.add_and_reg() and Tree.remove_and_unreg()

agrover:
I want to be able to have a method that winds up creating or removing an objectpath from the tree (e.g. a Create method that results in a representation of the just-created thing, enumerable with objectmanager.)

Anyway, I haven't found a way for that method to be *in* the Tree itself (we'd need a &mut Tree in the method's closure) so what I've done is have two trees - one with the objectpath supporting an interface with Create and Destroy, and then another Tree (not within the first) that we can get a &mut for -- with some Rc, RefCell and closuring in the first one's methods -- where the objects themselves live.

OK great. Now the Create and Destroy methods just need a way to add or remove a single objectpath after initialization. The current add/set_registered methods don't really work for this, hence the new add_and_reg and remove_and_unreg methods. But they have their own downsides in that the two pairs of methods should really not be mixed, and `add_and_reg` isn't as chainable (?) as `add` is.

Anyway, please let me know what you think! :smile: thanks.

### DIFF
--- a/src/methoddisp.rs
+++ b/src/methoddisp.rs
@@ -675,6 +675,20 @@ impl<M: MethodType> Tree<M> {
         self
     }
 
+    /// Add a Object Path to this Tree, and register it.
+    pub fn add_and_reg(&mut self, c: &Connection, p: ObjectPath<M>) -> Result<&mut Self, Error> {
+        try!(c.register_object_path(&*p.name));
+        self.paths.insert(p.name.clone(), Arc::new(p));
+        Ok(self)
+    }
+
+    /// Remove an Object Path from this Tree, and unregister it.
+    pub fn remove_and_unreg(&mut self, c: &Connection, p: ObjectPath<M>) -> &mut Self {
+        c.unregister_object_path(&**p.name);
+        self.paths.remove(&p.name);
+        self
+    }
+
     /// Registers or unregisters all object paths in the tree.
     pub fn set_registered(&self, c: &Connection, b: bool) -> Result<(), Error> {
         let mut regd_paths = Vec::new();


diwic:
The remove one looks weird, as it consumes an `ObjectPath<M>`, this is not correct. The remove function should take a name only. The problem is: what *is* right? I would have liked a `&Path<'a>` but that did not please the borrow checker, for reasons I fail to understand. So my current version uses a `T: Into<Path<'static>>`, but it bothers me that I can't get it to work the way I want it to, so I haven't pushed it.

diwic:
> Anyway, I haven't found a way for that method to be in the Tree itself (we'd need a &mut Tree in the method's closure) so what I've done is have two trees

I would have deferred the operation; keep a `Vec` of "things to do" which you add to during the method call. After the method has completed (i e, outside `Tree::handle`), you can just check your `Vec` and carry out the add/remove operations. (You might need to have that `Vec` inside a `RefCell` but that shouldn't be a problem.)

agrover:
Ah yes, a deferred list might do the trick. I'll play with my code some more and see if it lets me do that, and maybe get back down to the single Tree.

agrover:
`Path` is

agrover:
`Path` is a `Cow<'static, CStr>` so what about making the arg `p: &Path<'static>`? but I'd tend to have the name in a `&str` instead of a `&CStr`. The `Into<Path<'static>>` I guess handles the conversion to `Path` containing an owned CString, which I actually kinda like.

diwic:
Fair enough, `&Path<'static>` it is for now, then I can change it into `&Path<'something>` in case I figure out how to please the borrow checker better. Pushed now; so I'm going to close this one.

PR #42, opened 2016-04-22
add Tree::add_o_ref

agrover:
This lets us add an ObjectPath to a Tree we only have a &mut for.

### DIFF
--- a/src/methoddisp.rs
+++ b/src/methoddisp.rs
@@ -669,7 +669,7 @@ impl<M: MethodType> Tree<M> {
         }).collect()
     }
 
-    /// Add a Object Path to this Tree.
+    /// Add an Object Path to this Tree.
     ///
     /// Note: This does not unregister a path with the connection, so if the tree is currently registered,
     /// you might want to call Connection::register_object_path to add the path manually.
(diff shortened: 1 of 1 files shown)


agrover:
There's probably a better name than `add_ref`, but see https://github.com/agrover/froyo/blob/wip2/src/dbus_api.rs#L278. I only have a &mut reference to the tree, not ownership.

diwic:
FWIW, It's easy to get a `Self` from a `&mut self`:

     let t = mem::replace(&mut tree, Tree::new_fn());
     let t = t.add(path);
     mem::replace(&mut tree, t);

diwic:
That said, the above is a workaround for a legitimate use case. I'm not sure whether to:
 1) Ask you to have it return a `Arc<ObjectPath<M>>` instead of a `&mut self`, for consistency with other `add_*_ref` methods. If we do that `remove` must return an `Option<Arc<ObjectPath<M>>>` instead of a `Option<ObjectPath<M>>`.
 2) Keep it as it is but change the name. This feels more inconsistent with the rest of the tree API, so I'm leaning towards the first option.

Either way, you should add a comment as you were the one introducing comments for all public functions in the first place ;-)

agrover:
Updated commit. How's this look? Thanks!

diwic:
Looks good!

agrover:
Great! Say, might it be time for a release?

diwic:
Sure, why not? Done, enjoy!

PR #48, opened 2016-09-13
make cstring wrappers hashable for use in hashmaps

t-8ch:


diwic:
Looks good to me, thanks!

PR #49, opened 2016-09-17
Fix typo in service exmple in README

jonte:
Factory.interface() needs the Data parameter


diwic:
Ah, and it's in the Readme, so "cargo test" doesn't catch it. Thanks!

PR #54, opened 2016-10-31
Add tree-wide data item

agrover:
Hi diwic, would being able to attach data to the overall tree also be useful?

We're using a `Rc<Refcell<T>>` to share ownership across each objectpath of our "dbus context", but it seems a little nicer if we could hook tree-wide stuff to the tree, and save the objectpath spot for objectpath-specific stuff. Thanks in advance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ It then listens for incoming D-Bus events and handles them accordingly.
 let c = Connection::get_private(BusType::Session).unwrap();
 c.register_name("com.example.dbustest", NameFlag::ReplaceExisting as u32).unwrap();
 let f = Factory::new_fn::<()>();
-let tree = f.tree().add(f.object_path("/hello", ()).introspectable().add(
+let tree = f.tree(()).add(f.object_path("/hello", ()).introspectable().add(
     f.interface("com.example.dbustest", ()).add_m(
         f.method("Hello", (), |m| {
             let s = format!("Hello {}!", m.msg.sender().unwrap());
--- a/examples/adv_server.rs
+++ b/examples/adv_server.rs
@@ -27,6 +27,7 @@ struct Device {
 #[derive(Copy, Clone, Default, Debug)]
 struct TData;
 impl tree::DataType for TData {
+    type Tree = ();
     type ObjectPath = Arc<Device>;
     type Property = ();
     type Interface = ();
@@ -126,7 +127,7 @@ fn create_tree(devices: &[Arc<Device>], iface: &Arc<Interface<MTFn<TData>, TData
     -> tree::Tree<MTFn<TData>, TData> {
 
     let f = tree::Factory::new_fn();
-    let mut tree = f.tree();
+    let mut tree = f.tree(());
     for dev in devices {
         tree = tree.add(f.object_path(dev.path.clone(), dev.clone())
             .introspectable()
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -29,7 +29,7 @@ fn main() {
     let signal2 = signal.clone();
 
     // We create a tree with one object path inside and make that path introspectable.
-    let tree = f.tree().add(f.object_path("/hello", ()).introspectable().add(
+    let tree = f.tree(()).add(f.object_path("/hello", ()).introspectable().add(
 
         // We add an interface to the object path...
         f.interface("com.example.dbustest", ()).add_m(
--- a/src/tree/factory.rs
+++ b/src/tree/factory.rs
@@ -4,7 +4,6 @@ use super::objectpath::IfaceCache;
 use std::sync::Arc;
 use Interface as IfaceName;
 use {Member, Path, arg};
-use std::default::Default;
 use std::cell::RefCell;
 
 /// The factory is used to create object paths, interfaces, methods etc.
@@ -86,8 +85,9 @@ impl<M: MethodType<D>, D: DataType> Factory<M, D> {
     }
 
     /// Creates a new tree.
-    pub fn tree(&self) -> Tree<M, D> { Default::default() }
- 
+    pub fn tree(&self, data: D::Tree) -> Tree<M, D> {
+        super::objectpath::new_tree(data)
+    }
 }
 
 
--- a/src/tree/factory.rs
+++ b/src/tree/factory.rs
@@ -108,6 +108,7 @@ fn fn_customdata() {
     #[derive(Default)]
     struct Custom;
     impl DataType for Custom {
+        type Tree = ();
         type ObjectPath = Arc<u8>;
         type Interface = ();
         type Property = ();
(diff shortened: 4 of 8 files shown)


diwic:
I'm all for having a data attached to the tree in case you find it useful. But, did you run a "cargo test" on your commit to verify that you did not break any test cases? Because adding something to DataType seems like a potentially build-breaking change for examples and/or tests.

diwic:
> save the objectpath spot for objectpath-specific stuff

Side note: Not sure how this "consumes" the objectpath spot - you can always have, e g, a tuple `(Rc<RefCell<MyDBusContext>>, ObjectPathSpecificStuff)` as objectpath data :-)

agrover:
Yes I'm sure it breaks things since it's an API change. Since you seem amenable, let me go ahead and update the PR to fix examples/tests, and then I'll ping you again.

agrover:
@diwic ok, fixed docs and examples.

diwic:
The server example in the readme also needs fixing.

SnirkImmington:
> Yes I'm sure it breaks things since it's an API change. Since you seem amenable, let me go ahead and update the PR to fix examples/tests, and then I'll ping you again.

Please don't break the API. If there's one thing the dbus crate can do it's backwards compatibility, look at the deprecated `obj` crate still in the tree. You're making this change and not even deprecating anything or adding any documentation about why you're making it. Other crates depend on dus too :smile:

Could you make your new methods have different names? For example, leave `Factory::tree(&self)` the way it is, and add `Factory::tree_with_data(&self, data: D::Tree)`? You don't need to break everyone else's code for this change.

agrover:
@diwic: README updated

@SnirkImmington: yeah I was assuming we'd need to bump version to 0.5? That would allow 0.4 users to keep using it w/o change. I don't know if just adding a new constructor would do it since DataType needs another associated type, but happy to oblige if possible...

(I know semver says for x.y.z change x on API break, but for the moment dbus-rs seems to be using y for this.)

diwic:
@SnirkImmington, thanks for making your voice heard. Seems like I got my first conflict of interest here. 

Trying to figure out which way to go, I had a quick look at [dbus reverse dependencies](https://crates.io/crates/dbus/reverse_dependencies), only two crates are at 0.4 of DBus, the rest are at 0.3 or lower, so they're probably not interested in the "latest and greatest". The two crates at 0.4 are dbus-macros, which is affected (need to change its code in two places), and notify-rust which is not affected (still using dbus::obj). 

Certainly there can be more software (that's not on crates.io) that's affected, but a quick glance seems to indicate that it's not that many. Probably having a consistent API (instead of adding a separate `tree_with_data` method) weighs heavier. And the change is not that difficult to do for the user (just add `()` when you make a tree).

Merging this commit would make the next release 0.5, so it won't break people's builds until they want the latest and greatest for some reason.

agrover:
Yeah, it's tricky to judge because apps aren't necessarily on crates.io.

SnirkImmington:
> yeah I was assuming we'd need to bump version to 0.5?

I've been using the y part of the semver for backwards compatibility too. So many Rust crates are afraid, unwilling, or unable to go past `1.0.0` anyway. It just seems like bad practice to be willing to do for smaller changes that you probably could sit down and scratch heads over for a couple of days and integrate without breaking a very commonly used API method. 
It's important for library developers to show they are willing to be careful and thoughtful about their changes. Obviously at release 0.5 this isn't high-priority but as an upstream user I was surprised that a random breaking change could happen so quickly.

As a side note, would you be opposed to posting an example about where this data could be used and how it's useful? Looking at the changeset I get the feeling that all of the examples haven't been looked at and will break. 

I also don't understand the reason for adding data directly to the tree - would this be like if you had a dbus server hosting a collection of widgets, you'd plop your `Vec<Widget>` directly into your dbus server instead of as a variable alongside?

agrover:
You can see the gory details here: https://github.com/stratis-storage/stratisd/blob/master/src/dbus_api.rs#L47 (one detail: we're currently refcounting members of our context instead of the overall context, but this isn't relevant to the below explanation)

but in short, we need to change common data structures from within multiple methods. This involves a lot of `Rc<RefCell<>>` of course, and right now we're using the ObjectPath data member to each hold a reference. All objectpath datas point at the same thing.

It would be a little nicer if we could instead stick that in one common place that every method in the tree could still get to, and then use the per-objectpath data for info is actually objectpath-specific. Diwic's comment above about using a tuple is of course correct, or we could use per-method data to point at the global shared context, but I think the solution that closest models what we're actually trying to do is to have a per-tree data item.

diwic:
@agrover - having looked a bit at the code you pointed, perhaps another way of doing the same is to move `Rc<RefCell<Context>>` (or `Rc<RefCell<Engine>>`) into the methods? Like this:

    let context2: Rc<RefCell<Context>> = context.clone();
    f.method(CREATE_SNAPSHOT, (), move |minfo| create_snapshot(minfo, context2));

As a side note, I think it's slightly scary that my code is being used to destroy file systems. ;-)

diwic:
@SnirkImmington 
> So many Rust crates are afraid, unwilling, or unable to go to 1.0.0 anyway.

Indeed. Being in 0.x land is nice for the library developer but not so much for the library user. (And not even the libc crate is 1.0...) You're also the first to actively ask for more stability.

So yes, there are a few more harder things to tackle before I'm at 1.0. But maybe it makes sense to start planning.

agrover:
@diwic Yes, that's what was required back before MethodInfo changes in 0.4. Certainly still doable.

What exactly were the reasons for the big Tree rewrite in 0.4 that introduced MethodInfo? I didn't see it discussed in the commit log, but *I* saw its main advantage as a more straightforward way to get data in a method without requiring closing over an Rc value :smile:

diwic:
> What exactly were the reasons for the big Tree rewrite in 0.4

Better memory efficiency - now we can `Arc<Interface>` and put that same struct on many objectpaths.

 >  that introduced MethodInfo?

Well, in case you do use the same Interface and Methods for many object paths, it makes sense that it's easy to figure out *which* object path the method was called on. :-)

The associated data items on object paths, interfaces etc are more of an experiment. It just seemed like something useful to add in this scenario; if one object is modelled as one object path, it would be helpful if the object path could contain some kind of link to that value. Whether that means owning the value, shared ownership, a key to lookup in a map somewhere, etc...would be up to the application using the tree.

diwic:
Ok, and the decision is to merge. All things considered, it's an easy change to follow along with (just find/replace `.tree()` with `.tree(())`). 

But @agrover - if you have more changes that might break the API please have them coming within the next days/weeks or so, because after 0.5 I might be a little more hesitant to add them given @SnirkImmington s request to have the crate be more stable API wise.

diwic:
I've also added a "v1.0" milestone and a few bugs that I want to see done before I'm willing to call it 1.0 for dbus-rs. It would be nice to see the libc crate go to 1.0 too, though.

agrover:
Thanks. I'll be sure to raise any pre-1.0 issues ASAP -- none at the moment.

PR #59, opened 2016-11-13
Copy enum docstrings from dbus/dbus-shared.h

nicholasbishop:


diwic:
Hey @agrover - since you work for Red Hat, do you know if there's some legal sign-off required for copy-pasting a few strings this way? I mean, I have a hard time even imagining Red Hat would go after us for doing so, but better ask one time too much than one time too little...

agrover:
IANAL but you should not just apply this b/c RH is nice. RH could be bought by litigious jerks, who knows.

Getting permission would involve getting RH or the original author Havok Pennington, who may have kept original copyright to this work, to agree to release it under Apache 2 instead of Academic Free License v2.1 (which is very close to Apache2 but *not identical*). Or maybe they actually are already compatible but can't know for sure without in-depth legal analysis. Is it really worth it for these little bits???

I would recommend either adding a link to the original file if people want to see the comments, or rewriting the docstrings in your own words, rather than merging.

diwic:
Ok, so I'm going to close this PR without merging, based on agrover's recommendation. Sorry!

PR #61, opened 2016-12-24
Use metadeps to specify pkg-config dependencies declaratively

joshtriplett:
This makes it easier for distribution packaging tools to generate
appropriate package dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,10 @@ build = "build.rs"
 libc = "0.2.7"
 
 [build-dependencies]
-pkg-config = "0.3"
+metadeps = "1"
 
 [dev-dependencies]
 tempdir = "0.3"
+
+[package.metadata.pkg-config]
+dbus-1 = "1.6"
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,5 @@
-extern crate pkg_config;
+extern crate metadeps;
 
 fn main() {
-    pkg_config::find_library("dbus-1").unwrap();
+    metadeps::probe().unwrap();
 }


diwic:
Looks good to me in general: I understand metadeps is new and has not had much testing, have you tried it on anything non-Linux? The reason I'm asking is because IIRC the pkg-config dependency was originally added by someone who had problems compiling dbus-rs on some BSD variant, and I don't want to break that.

joshtriplett:
metadeps uses the pkg-config crate; it'll work anywhere pkg-config does.  I've submitted PRs to projects that test on several non-Linux OSes via CI, and those PRs have passed CI on all platforms without any issues.

diwic:
I tried it, but now compilation fails:

    error: attributes on non-item statements and expressions are experimental. (see issue #15701)
       --> /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/error-chain-0.7.1/src/lib.rs:419:9
        |
    419 |         #[cfg(feature = "backtrace")]
        |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |
        = help: add #![feature(stmt_expr_attributes)] to the crate attributes to enable

This is with `rustc 1.13.0-nightly (d0623cf7b 2016-09-26)`

I'll try upgrading rust to see if that helps.

diwic:
Okay, upgrading to rustc 1.14 stable helped. Still, the amount of build dependencies increased by a factor of three (I think) by adding metadeps as a dependency - maybe you can consider turning some features off, e g, do we really need backtrace?

joshtriplett:
@diwic Working on that.  I've already gotten a change into error-chain which will make it work back to Rust 1.10, and I plan to turn off all the features of error-chain in metadeps 1.0.2.

joshtriplett:
@diwic Yeah, with backtrace disabled, error-chain has no other dependencies, so metadeps only adds itself, error-chain, toml, and rustc-serialize.

joshtriplett:
Done: current metadeps (1.0.2) now disables backtrace, and current error-chain (0.7.2) now runs on Rust 1.10 with backtrace disabled, so you should no longer need Rust 1.14 installed.

diwic:
Thanks!

PR #75, opened 2017-03-01
Fix compilation errors on platforms with unsigned chars

jFransham:
The cc I'm using for cross-compilation uses unsigned chars, which means this cast to `i8` fails. This PR just casts to `c_char` instead.


diwic:
Thanks! Sorry for breaking your build.

diwic:
@jFransham 
 This was a 0.5.0 -> 0.5.1 regression, right? So you'll need a 0.5.2 release as well?

jFransham:
I don't know, I hadn't used this library before yesterday. I can pull from git for now (I already have my local branch in the `Cargo.toml`), but I'm sure there's no harm in putting out a release for this.

PR #76, opened 2017-03-03
Add asynchronous send_with_reply() function

antoyo:
Hello.

This pull request add the `send_with_reply()` method for an asynchronous method call.

I wonder how errors should be treated. Currently, a out of memory error returns `Err(())`.

Also, I wonder if I should call `dbus_pending_call_get_completed()` before `dbus_pending_call_set_notify()` in case the response is already received.
If this can happen, the notify callback won't be called.
I don't know if this can happen.

Finally, I assume that the response from `dbus_pending_call_steal_reply()` won't be null, since it is only called when notified.

### DIFF
--- a/examples/async_client.rs
+++ b/examples/async_client.rs
@@ -1,0 +1,19 @@
+extern crate dbus;
+
+use dbus::{Connection, BusType, Message};
+use dbus::arg::Array;
+
+fn main() {
+    let c = Connection::get_private(BusType::Session).unwrap();
+    let m = Message::new_method_call("org.freedesktop.DBus", "/", "org.freedesktop.DBus", "ListNames").unwrap();
+    c.send_with_reply(m, |result| {
+        let r = result.unwrap();
+        // ListNames returns one argument, which is an array of strings.
+        let arr: Array<&str, _>  = r.get1().unwrap();
+        for name in arr { println!("{}", name); }
+    }, 2000).unwrap();
+
+    for _ in c.iter(0) {
+    }
+}
+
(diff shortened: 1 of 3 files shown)


diwic:
Hi! And thanks for the interest in improving the library!

Having looked through the patch, I'm okay with the idea as such, but I'd rather see that you did not call the callback directly from the C function; this is mainly for panicking reasons, we need to handle the callback function panicking in a clean way.

However, this requirement requires the patch to be rewritten. As an alternative, what if we just kept a list of `Box<Fn<_>>` and their reply serials, internally checking this list when we process our list of pending_items?

diwic:
As for your questions:

 > I wonder how errors should be treated. Currently, a out of memory error returns Err(()).

As a general guidance, this library panics on out-of-memory errors, because that's what Rust does. Other errors should be handled properly.

 > Finally, I assume that the response from dbus_pending_call_steal_reply() won't be null, since it is only called when notified.

I believe this is a wrong assumption since the callback is also called on timeout, but anyhow, this is likely irrelevant in case the patch is rewritten to not use the dbus_pending_call API.

diwic:
Since you didn't reply, I went ahead and implemented my own version instead. Here's how you'll write it now:

    let c = Connection::get_private(BusType::Session).unwrap();
    let m = Message::new_method_call("org.freedesktop.DBus", "/", "org.freedesktop.DBus", "ListNames").unwrap();
    let reply = c.send_with_reply(m, |result| {
        let r = result;
        let _: ::arg::Array<&str, _>  = r.get1().unwrap();
    });
    c.iter(1000).with(reply).count();

PR #91, opened 2017-07-02
Move crate folders to top and add virtual workspace

albel727:
It's much more convenient to build/test with `cargo build --all` (and eventually even specifying --all won't be necessary). 

`path` dependencies allow testing local changes across several crates at once. 

Also setting workspace enables common `Cargo.lock` for all crates in the top directory, so gitignoring nested `Cargo.lock`s is not necessary.


diwic:
I'm fine either way, so if it makes you happy :-)

albel727:
Yeah, it rather does. It was becoming a bit tedious to move things back and forth for testing and committing. Thanks!

Issue #72, opened 2017-01-29
tokio integration


diwic:
It would be nice if it was easier to use this dbus library in an async fashion.

brson:
@antoyo, @zeenix and I spent some time hacking on a proof of concept of tokio integration. We managed to create a basic example client, started hacking on an example server, and then began feeling a bit over our heads w/r/t dbus-rs internals. I hope that, at the least, this work can serve as guidance toward the real solution.

Here's the branch: https://github.com/brson/dbus-rs/tree/tokio

The entire implementation is contained in [tokio.rs](https://github.com/brson/dbus-rs/blob/tokio/src/tokio.rs). There is a [client example](https://github.com/brson/dbus-rs/blob/tokio/examples/tokio_smoke.rs) and [server](https://github.com/brson/dbus-rs/blob/tokio/examples/tokio_server.rs).

The comment at the top of tokio.rs describes the integration strategy, which is derived from [tokio-curl](https://github.com/tokio-rs/tokio-curl/blob/master/src/unix.rs).

This patch integrates directly into dbus-rs for simplicity, but it would be just as well as a standalone crate.

One unfortunate thing about this integration is that it needs to know directly about tokio-core and mio in order to communicate information about the fds that come out of dbus. So it's not completely abstracted from the underlying event loop, and will need specific integration into future tokio event loop implementations. I figure solutions there will shake out in time.

Since we didn't make much progress on the server implementation, we just wrote the server example to demonstrate what might be a plausible API. In the client code we took a strategy of encapsulating Connection in TokioConnection; for the server we might similarly encapsulate it in TokioServer, or possibly give TokioConnection a method that returns a server future of some kind, which can then be used to drive an event loop (this seems like it may be more consistent with the existing Connection design).

The big thing we got stuck on was understanding the parameterization of `Factory` over a function abstraction, and in particular what `MTFn` does. I suspect we need some serious guidance at this point on how to proceed in a way that is congruent with the existing design.

FYI, I don't personally expect to continue working on this, but I'm hopeful @antoyo and @zeenix can continue in collaboration with @diwic.

diwic:
[...]

PR #92, opened 2017-07-03
Fix poll hanging

albel727:
Fix as per https://github.com/diwic/dbus-rs/issues/72#issuecomment-312553089

On top of PR #91

### DIFF
--- a/dbus-tokio/src/adriver.rs
+++ b/dbus-tokio/src/adriver.rs
@@ -133,7 +133,7 @@ impl ADriver {
                     if let Some(r) = r { r.send(m).unwrap(); }
                     else { self.send_stream(m) }
                 }
-                ConnectionItem::Nothing => return,
+                ConnectionItem::Nothing => unreachable!(),
                 ConnectionItem::Signal(m) => self.send_stream(m),
                 ConnectionItem::MethodCall(m) => self.send_stream(m),
             }
@@ -168,7 +168,9 @@ impl Future for ADriver {
             if ur.is_readable() { w.need_read() };
             if ur.is_writable() { w.need_write() };
         };
-        self.handle_items(items.unwrap_or(cc.iter(0)));
+        if let Some(items) = items {
+            self.handle_items(items);
+        }
         Ok(Async::NotReady)
     }
 }


diwic:
Hmm, so the only thing that makes me a bit uneasy about this is that I actually had it this way, but decided to change it (in b6c02346 )

But maybe that was an incorrect fix for something you fixed correctly with the edge/level stuff?

diwic:
> `iter(0)` with `read_write_dispatch()` is not needed,
and removing it and relying on `watch_handle()` and
`iter(None)` with pure `dispatch()` fixes the bug.

Hmm, this is not what the patch does. There is no `iter(None)` AFAICS...?

albel727:
>But maybe that was an incorrect fix for something you fixed correctly with the edge/level stuff?

Maybe it was that. `iter(0)` just made some more progress with `read_write_dispatch()`, so it hid the fact that edge wakeups weren't always enough to read everything. With level triggering, everything is read with `dbus_watch_handle()` and dispatched with `iter(None)` by the `ADriver`, as you can test yourself.

>Hmm, this is not what the patch does. There is no iter(None) AFAICS...?

Oh, I meant it's implicit in `items = Some(cc.watch_handle(w.get_ref().0.fd(), flags));`. It returns `ConnectionItems { c: self, timeout_ms: None, handlers: Vec::new() }` which is what `iter(None)` would have returned, if it existed.

In fact there's a problem with that lack of `iter(None)` and `items` being conditional on having watched sockets too. Under some circumstances (like exceeding `dbus_connection_set_max_received_size()`/`dbus_connection_set_max_received_unix_fds()` limits), libdbus may disable both reading and writing watches, and then there wouldn't be `fds` to iterate over, and then there wouldn't be `items` set, and then there wouldn't be `handle_items()` called to process `WatchFd` if libdbus decides to re-enable reading. 

Another part of the problem is, `iter(0)` or, somewhat better, unconditional `iter(None)` at the end of `ADriver::poll()` won't help with that either, since, with events from watches disabled, nothing will poll the `ADriver` future anymore, so `WatchFd` would remain unprocessed anyway.

I'm going to try to think in my free time, how to solve that elegantly enough, as it may be necessary to sufficiently change the architecture of dbus-tokio. Meanwhile I'm rather sure that this PR brings dbus-tokio as close to being reliable, as possible, if you forget the incoming queue overflow problem, which exists with or without it anyway.

albel727:
Well, there's a lot of things that need to be considered, really. For instance, how things would work with `ADriver::modify_watch()` at all. For libdbus mostly doesn't change watches besides initial addition, and so everything appears to work now. But if it does, then we'll create a new `PollEvented`, and we'll do so in `handle_items()` after `fds` iteration, which calls `need_read()`/`need_write()`/`poll_ready()`. And so `ADriver` never registers its interest in the new `PollEvented`. And the old `PollEvented` is dropped, so `ADriver` will never be polled again.

albel727:
If there were `iter(None)`, I'd just solve that problem by unconditionally iterating over it in `ADriver::poll()`, and calling `need_read()`/`need_write()` in `modify_watch()`. Of course, there's no need to call `modify_watch()` in `AConnection::new()` manually then, and it would probably panic for being called outside of a Task context too.

That still won't solve the problem of completely disabled watches halting `ADriver` forever though.

diwic:
Hrm, so I merged it, and now my "cargo test" hangs (aconnection_test works, astream_test hangs).

albel727:
Ugh. And here I had to disable it while investigating `aconnection_test()` and missed a regression. Let me investigate.

diwic:
Okay, so what the `iter(0)` does (at least), is that it catches messages that are already in *our* pending_items queue. If this does not happen, all that happens is that the first fd is not ready, and since there is no `iter(0)`, the initial messages put into the queue are never processed.

diwic:
I just pushed a commit that makes us call dbus_connection_dispatch, although in a quite ugly manner :-)

Does this fix sound viable to you or did I destroy some other case?

albel727:
Oh, but of course. It's almost the same problem that I've tried to fix - exhausting socket without dispatching all messages. 

Only now the socket is exhausted by `dbus_bus_get_private()`, which registers unique name with the daemon and reads in everything in process, including the `NameAcquired` signal that we wait for, but doesn't dispatch it. So now the socket is empty and so `items` is None, and so we never run `iter()` to dispatch what's already in the queue. And since it all happens before we obtain the connection, not even `wakeup_main/dispatch_status` callbacks could have been of help.

I'm afraid `iter(None)` has to become a thing to properly fix this. Some access to `dbus_connection_dispatch()` that is not via `dbus_connection_read_write_dispatch()` is sorely needed. Maybe `Connection::iter_without_io()` or something?

albel727:
Aaand now I see your message about `dispatch()`. Let me check out your commit. :)

albel727:
>self.handle_items(items.unwrap_or(cc.watch_handle(-1, 0)));

Haha! That's exactly what I wrote when I checked that unconditional `iter(None)` is really what would work. Yeah, I expect that to work. 

Well, now that `iter(None)` is there, I can fix the `PollEvented` recreation bug too.

PR #101, opened 2017-09-06
tokio_client example

jacekchmiel:
Hi diwic,

I've created a little more advanced tokio-client example, that simultaneously calls methods and processes signals. I think it could be really useful for others (It would've been for me in the first place). 

I'm really new to rust and tokio so the code may be far from ideal. I'd be more than happy to improve this example if needed, just give me the directions.


diwic:
Hi! I think this example looks really good! (Given my own limited experience with Tokio.)

I'll be happy to merge it.

Issue #102, opened 2017-09-11
Tokio: Method arguments for servers


manuels:
I have a problem figuring out how to implement a tokio dbus server with method arguments.

The example states `f.amethod("Hello", (), move |m| { ... }` and apparently the `()` token is the argument list, right?
The documentation states

```
fn amethod<H, R, T>(
    &self, 
    t: T, 
    data: D::Method, 
    handler: H
) -> Method<MTFn<ATree<D>>, ATree<D>> where
    H: 'static + Fn(&MethodInfo<MTFn<ATree<D>>, ATree<D>>) -> R,
    T: Into<Member<'static>>,
    R: 'static + IntoFuture<Item = Vec<Message>, Error = MethodErr
```
and
```
type Method: Debug + Default
```

Still, I have no idea, how to add a function to the interface that has arguments...
A little help here would be appreciated!

diwic:
Hmm, it seems like you are confusing `dbus::tree::Method` and `dbus::tree::DataType::Method`?

The `()`, which refers to `dbus::tree::DataType::Method` is not an argument list. It's some extra data you can store in-tree if you wish. You can probably ignore it for now.

Arguments are added like this:

    f.amethod( /*... */)
        .inarg::<i32,_>("foo")
        .inarg::<bool,_>("bar")
        .outarg::<&str,_>("baz")

In the example above, the method takes two arguments (foo and bar) and returns one output argument (baz).

PR #103, opened 2017-09-17
Add Method::inarg() to examples

manuels:
Sorry, I lost track of the issue #102.
I added some Method::inarg() calls to the examples to make it clear to new users how to use arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,10 @@ let f = Factory::new_fn::<()>();
 let tree = f.tree(()).add(f.object_path("/hello", ()).introspectable().add(
     f.interface("com.example.dbustest", ()).add_m(
         f.method("Hello", (), |m| {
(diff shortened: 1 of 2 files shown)


diwic:
Hi and thanks for improving the examples! 

However, I'd really like proper error handling w r t parsing the input arguments from the client - otherwise it'll be too easy to make the server panic. 

I fixed this up for the easy example (see https://github.com/diwic/dbus-rs/commit/10c616d95984558d75cab4057dc50ec7cfbb801e ). Could you help me with doing the same for tokio-server? I'm just lost in how to write "?" or "early return" tokio style.

manuels:
Yeah, tokio is a little bit of a pain in the ass in that context.
The situation will get better once [future-await](https://github.com/alexcrichton/futures-await) has landed.
I would recommend to wait until that has happend. Otherwise, the example code would become too complicated.

Issue #105, opened 2017-10-04
`codegen`d objects are not introspectable


jonnymatts:
I'd like to use D-Feet to debug my dbus service, but it requires that services be introspectable.

At the moment, my workaround is to copy-paste all required interfaces into one xml file, that I use for code generation and also `include_str!` in my implementation of `org.freedesktop.DBus.Introspectable`. 

I wonder how difficult it would be to automatically implement `Introspectable` for all services, either by including the xml that was used to generate them, or by regenerating it from other metadata.

diwic:
If you are using the `tree` module, just call `introspectable()` on the object paths you want to allow introspection on.

albx79:
My bad for not reading the docs! 

Indeed, calling `introspectable()` does the trick. Perhaps it's worth adding it to the codegen examples, as it's quite easy to get lost in the code and forget how the API works.

diwic:
Ok, feel free to submit a PR for better docs - as you probably know better than I do where they should have been to make it easier for you.

PR #106, opened 2017-10-05
Clarify server-side usage

albx79:
as discussed in issue #105 , I think this small change, with a link to the main README, will make this section slightly clearer.


diwic:
Thanks!

PR #111, opened 2017-11-24
Remove the metadeps dependency and bump version

Eijebong:
Just use the pkg_config package directly instead of using the metadeps
one which requires an old version of toml.

cc @nox

### DIFF
--- a/libdbus-sys/Cargo.toml
+++ b/libdbus-sys/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "libdbus-sys"
-version = "0.1.1"
+version = "0.1.2"
 authors = ["David Henningsson <diwic@ubuntu.com>"]
 
 description = "FFI bindings to libdbus."
@@ -13,7 +13,4 @@ categories = ["os::unix-apis", "external-ffi-bindings"]
 build = "build.rs"
 
 [build-dependencies]
-metadeps = "1"
-
-[package.metadata.pkg-config]
-dbus-1 = "1.6"
+pkg-config = "0.3"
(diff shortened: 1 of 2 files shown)


diwic:
Could you elaborate?
I believe the metadeps dependency was added for a reason, is that reason no longer relevant? 

cc @joshtriplett

nox:
toml 0.4 depends on serde 1.0, and some people don't want to have to compile serde for their project. This means that metadeps is staying on toml 0.2 to avoid the serde dependency, and any project that uses toml 0.4 and metadeps now need to compile 2 toml parsers.

metadeps exist to be able to specify system dependencies in a declarative way, but AFAIK no other tool was ever made to read them anyway, so I don't think that's worth the compiling 2 toml parsers for no reason.

diwic:
Hmm, so I believe @joshtriplett had some idea that this would be helpful for Debian packaging tools. I'm a little hesitant about removing it just like that.

Also, if people have a problem with metadeps, it should be fixed in metadeps: e g by adding some feature that controls what version of the toml crate that metadeps depends on?

Eijebong:
Is there any way to move this forward ?
I don't see how that would help debian packaging tools btw, those are working with patches anyway. Changing the required version or the version name for a lib would be the same effort if you have to change the Cargo.toml file or the build.rs one

diwic:
Ok, I've filed https://github.com/joshtriplett/metadeps/issues/9 - let's see what that gives us.

diwic:
@nox 

 > no other tool was ever made to read them anyway

So I believe the tool that uses this info is [debcargo](https://crates.io/crates/debcargo) ?

PR #113, opened 2017-12-05
libdbus, dbus: Add connection_get_is_connected to get whether current connection is open or not.

kubo39:


diwic:
Sure, why not! Merged. Thanks for the patch.

PR #116, opened 2018-01-03
Provide way to register connection with bus

cpick:
Many bus connections require registration before methods can be called.
`Connection::get_private` does this automatically
(via `dbus_bus_get_private`), but `Connection::open_private`
(`dbus_connection_open_private`) does not.

Add `Connection::register` (which calls `dbus_bus_register`) so
`Connection`s created with `open_private` can register.


diwic:
Looks good to me. Thanks!

PR #119, opened 2018-01-29
[META] Fix lixensing

### LINKED ISSUE #117: include LICENSE into sub-crates and add MIT text license
* Seems MIT text license is missing
* None of sub-crates contain any licenses so archives on crates.io don't have them and this needs to be fixed

1wilkens:
Add proper licenses to the repository root and crate subfolders.
I am not sure about the `license owner` text in the files, as the crates contain commits from other authors but as the name is the only one in `Cargo.toml` I think it should be fine for now.

This should fix #117.


diwic:
Hi and thanks for the PR!

I guess it would be more fair to say "David Henningsson <diwic@ubuntu.com> and other contributors".

Another question; if you made  libdbus-sys/LICENSE-APACHE a symlink to ./LICENSE-APACHE instead of making five copies of the file, would that work or will cargo get confused? (Same for LICENSE-MIT and the other subdirectories)

1wilkens:
I'll fix the wording ;)

I thought about symlinking or at least removing the licenses from the repo root (although that would hurt discoverability and tools like GitHub license detection).

Actually, I am not sure if `cargo` detects these files at all, as the `license` field on `crates.io` just uses the value from `Cargo.toml` afaik.

1wilkens:
Pushed!
I also think I tried to be a little to pedantic by estimating the project start of each subcrate to include in the licenses so now I just settled on 2014-2018 everywhere.

diwic:
Merged. Thanks for the patch!

PR #121, opened 2018-02-01
[dbus-codegen] Add UnixFd to supported ArgTypes in generate::xml_to_rust_type

### LINKED ISSUE #120: [dbus-codegen] `dbus-codegen` panics on handles in arguments
When trying to use `dbus-codegen` to generate bindings for the [logind Manager API](https://www.freedesktop.org/wiki/Software/systemd/logind/) I noticed it panicking if an argument with the type "h"/Handle was encountered.

The panic message given was
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: StringError("Unknown character in signature Ok(UnixFd)")'`
```

It does not matter whether the argument is "in" or "out" as the following example quickly shows:
```xml
<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
<node>
    <interface name="org.example">
        <method name="SomeMethod">
            <arg name="arg0" direction="in" type="h" />
            <arg name="result" direction="out" type="h" />
        </method>
    </interface>
</node>
```

I don't know if this error is based on any implementation specifics which prevent `UnixFd`s from appearing in arguments or if this is just not yet implemented. In any case the call to `unwrap` should probably replaced with an `expect` or handled otherwise.

I'll check out the code and see if I can whip up aquick PR.

EDIT: Okay although the backtrace was not helpful I think the `unwrap` occurs directly in `main` with the error being bubbled up from `generate`. Panicking should be fine here which leaves the question if and how the error message could be improved.
Also as I see [here](https://github.com/diwic/dbus-rs/blob/639ddbcf78fa9f1fb7e0a0a9d56cf12aef9405bf/dbus-codegen/src/generate.rs#L156) the issue seems indeed to point to a missing implementation.
@diwic Any pointers on how I could approach fixing this?

1wilkens:
So this is my attempt at adding UnixFd to generate::xml_to_rust_type.
Really I just copied the above code and added the 'static lifetime, I think this is correct as we deal with a dbus type?

Fixes #120.

### DIFF
--- a/dbus-codegen/src/generate.rs
+++ b/dbus-codegen/src/generate.rs
@@ -126,6 +126,7 @@ fn xml_to_rust_type<I: Iterator<Item=char>>(i: &mut iter::Peekable<I>, out: bool
         Ok(ArgType::Int64) => "i64".into(),
         Ok(ArgType::UInt64) => "u64".into(),
         Ok(ArgType::Double) => "f64".into(),
+        Ok(ArgType::UnixFd) => "dbus::OwnedFd".into(),
         Ok(ArgType::String) => if out { "String".into() } else { "&str".into() },
         Ok(ArgType::ObjectPath) => if out { "dbus::Path<'static>" } else { "dbus::Path" }.into(),
         Ok(ArgType::Signature) => if out { "dbus::Signature<'static>" } else { "dbus::Signature" }.into(),


diwic:
dbus::OwnedFd has no lifetime parameter, so you shouldn't have one regardless of direction.

1wilkens:
Oops, I could have caught that, sorry! Will be fixed in a minute..

diwic:
Hmm, the commit now has a "fail.xml" without being referenced from any other file?

1wilkens:
Third times the charm! That was just my test file which I posted in the linked issue as well..

PR #126, opened 2018-02-25
Add links attribute

albel727:
This add `links = "dbus"` key to `libdbus-sys`, so as to be able to override linkage with cargo/config.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#overriding-build-scripts

This is a recommended practice, and is the easiest way to customize linking, which is helpful e.g. when cross-compiling (which metadeps/pkg-config doesn't readily handle). Unless there's a dedicated `[target.<arch-triple>.dbus]` config section in cargo/config, this key has no effect, so there are no contra-indications to adding it.

Someone on IRC mentioned having cross-compilation issue with this crate, so I thought to quickly fix that.

This PR also bumps libdbus-sys version to 0.1.3.

### DIFF
--- a/libdbus-sys/Cargo.toml
+++ b/libdbus-sys/Cargo.toml
@@ -11,6 +11,7 @@ keywords = ["D-Bus", "DBus", "libdbus"]
 license = "Apache-2.0/MIT"
 categories = ["os::unix-apis", "external-ffi-bindings"]
 build = "build.rs"
+links = "dbus"
 
 [build-dependencies]
 metadeps = "1"
--- a/libdbus-sys/Cargo.toml
+++ b/libdbus-sys/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "libdbus-sys"
-version = "0.1.2"
+version = "0.1.3"
 authors = ["David Henningsson <diwic@ubuntu.com>"]
 
 description = "FFI bindings to libdbus."


diwic:
Hi, nice to see you back again!

I wasn't familiar with the `links` attribute, but I was wondering if not `dbus-1` is the proper name to use here, since that's the name of the library as well as the pkg-config file?

albel727:
Thank you! :)

Well, the text itself is meaningless, it's just there for user to customize linkage and could as well be "banana" as long as other crates don't use it to mean a different library. 

Out of curiosity, I've passed over the most popular FFI crates on crates.io (excluding winapi), and they either don't use it altogether or write something unrelated there. Three out of three examples I stumbled upon of `links` usage by different developers didn't actually refer to library by pkg-config name, and sometimes not even by the actual file name.

E.g. this says "links = glib", but actually links to "glib-2.0".
https://github.com/gtk-rs/sys/blob/master/glib-sys/Cargo.toml#L31

This says "links = ffmpeg", but actually links to a whole bunch of libraries: "avcodec", "avdevice", etc.
https://github.com/meh/rust-ffmpeg-sys/blob/master/Cargo.toml#L5

This says "links = freetype", but links to "freetype2" with pkg-config, though despite being so-named in pkg-config the library file is actually "freetype".
https://github.com/PistonDevelopers/freetype-sys/blob/master/Cargo.toml#L6

Also if dbus library name contained dot, like "dbus-1.0", it would have conflicted with TOML section header parsing, but luckily it's not a concern for us, though I guess it was a concern for the glib example above.

So I gather that nobody cares to set links key precisely after library name, and ultimately it's just user typing convenience and maybe collision avoidance reasons. For the latter it seems it would be better if links key was somehow unique to the crate, like crate name minus "-sys" suffix, e.g. "libdbus", which also seems to be the general naming schema that people above used.

I don't have a preference here, so choose whatever you wish.

Issue #115, opened 2017-12-29
Add object path to the same tree at Runtime


LinkTed:
Hi it is me again,
it is possible to add a object path to the same tree at runtime?
My problem is that a dbus method create a new object path and I want to insert in the same tree, because the new object path have to access the data of the old object path(which created the new object path).
It can be done if it is possible to make the MethodInfo in the call handler mutable.
For example:
```rust 
factory.method("Method", (), move |m | { // if it is possible to make m mutable
...        
}
...
```

English is not my mother tongue: I hope you understand what i mean.

diwic:
No, you don't get a mutable tree inside a method handler. (This would allow you to delete whatever method you're currently calling, which would be unsound.) 

I should probably write an example for this as it is slightly tricky. You need to write a loop like:

     let mut new_path = None;
     factory.method("Method", (), |m| { new_path = Some( /* new path */); })
     /* ... */
     loop {
        if let Some(p) = new_path.take() { tree.add(p); }
        if let Some(msg) = connection.incoming(1000).next() {
            let r = tree.handle(msg);
            /* And send msgs in r */
        }
     }

LinkTed:
Yes, it works. :)
THANK YOU VERY MUCH.

Issue #52, opened 2016-10-11
Handling adding and removing object paths from a single tree


agrover:
Hiya,

We're still stuck on the issues raised by #40. We want the good stuff a tree provides (introspection, object manager, method dispatch) and also be able to add and remove object paths. This is important because we want to have our DBus interface create and remove objects and interfaces at runtime, and it's currently not possible, although we keep coming back to this as a natural way to want to design an API.

One easy thing would be to Add `Tree::add_o_path()` to tree/objectpath.rs `Tree`. This makes it possible to add an objectpath to the tree from a tree's method (although this requires an `unsafe` conversion of `&m.tree` to a `&mut m.tree`, it's a least possible) (This should not cause issues as long as it's always a different ObjectPath we're adding or removing than the objectpath/interface our method was invoked on)

Thinking longer-term, I'm wondering how hard it would be to make MethodInfo::tree a mutable ref (I tried it, `'a`s blossomed everywhere and I gave up, but you might have more luck) Or, I noticed that the only users of MethodInfo::tree are introspection and managed_objects, that call `tree::children()`, which then filters for children within the object path -- could there be a more distributed method of iterating and looking up object paths that would not require the entire `&Tree`? If so then an objectpath's method could be called without referencing the tree, and enclosing `Tree` in `Rc<RefCell>>` would let us close over and modify the tree without `unsafe`.

cc @tasleson @trgill @mulkieran

agrover:
Proposed implementation of tree/objectpath.rs Tree add_o_ref:
```
    pub fn add_o_ref<I: Into<Arc<ObjectPath<M, D>>>>(&mut self, s: I)
                                                     -> Arc<ObjectPath<M, D>> {
        let m = s.into();
        self.paths.insert(m.name.clone(), m.clone());
        m
    }
```

diwic:
Well, I'm not happy about sending a `&mut Tree` in, because if you could, you could remove all object paths, which could include the one you're currently calling a method on. That Rust protects you from being able to write that code is more a good than a bad thing IMO.

I believe the way to do this is in two steps: The method would just add an "add" or "remove" instruction in a TODO list. Back outside `tree::run` you have full mutable access to the tree, where you can carry out the instruction.

If you want to wait to send the reply back to the client until after the instruction has been fully carried out, you can also store the `method_return` message in the TODO list (and return an empty vec back in the method handler).

diwic:
I believe your suggested `add_o_ref` can be done today as:

[...]

PR #127, opened 2018-03-10
dbus: Create message from `MethodErr`

sbwtw:
new API: `method_error` in Message

Create a new `Message` from `MethodErr`

### DIFF
--- a/dbus/src/message.rs
+++ b/dbus/src/message.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::{fmt, mem, ptr, ops};
 use super::{ffi, Error, MessageType, Signature, libc, to_c_str, c_str_to_slice, init_dbus};
 use super::{BusName, Path, Interface, Member, ErrorName, Connection, SignalArgs};
+use super::tree::MethodErr;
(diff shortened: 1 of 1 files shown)


diwic:
Hmm, in general this is done internally in `Tree::handle`. I don't mind adding a public API for it, just curious why you need it?

diwic:
Also, I would prefer if `MethodErr` had a `to_message_reply` method instead - so the tree module depends on the "base" stuff in the root but not the other way around.

sbwtw:
I need dynamic add or remove object path from a `Tree`, and for insert, the new ObjectPath should be return. Like https://github.com/diwic/dbus-rs/issues/52 or https://github.com/diwic/dbus-rs/issues/115 . 
So I write a new struct and implements `MsgHandler` to do this. in DBus method call, return a empty `Result`, and in handler, create and insert new ObjectPath, then write this path to real HandlerResult. 
When ObjectPath insert failed, need to return a MethodError, so I need a easy way to convert this MethodError to a simple Message

diwic:
Can't claim I fully understand what you want to do, but anyhow, I've added a `MethodErr::to_message` that I hope solves your problem.

PR #129, opened 2018-04-01
WIP: Mostly ported to futures 0.2

rubdos:
Preliminary version of futures 0.2 support.
Not everything has ported; what's left is open for discussion.
Tokio hasn't officially announced their futures 0.2 support yet, [and it's not yet ready](https://github.com/tokio-rs/tokio/issues/251).

This was an experiment for me to see how far they were, and how simple dbus-rs could be ported.

I'm not sure how `PollEvented` works exactly; they do not have the `need_read` and `need_write` methods anymore, and I'm not sure what the replacement was.

---

To be done:

- [ ] Fix the `PollEvented` stuff. I'd like some input here.
- [ ] Patch the examples
- [ ] When Tokio has official support, pin the versions
- [ ] Make a breaking version change (version 0.3.0 or something)?
- [ ] Testing!

### DIFF
--- a/dbus-tokio/Cargo.toml
+++ b/dbus-tokio/Cargo.toml
@@ -12,13 +12,12 @@ categories = ["os::unix-apis", "api-bindings", "asynchronous"]
 
 [dependencies]
 dbus = { path = "../dbus", version = "0.6" }
-futures = "0.1.12"
+tokio-reactor = { version = "0.1.1", features = ["futures2"]}
+futures = "0.2.0-beta"
+futures-channel = "0.2.0-beta"
+futures-executor = "0.2.0-beta"
 mio = "0.6.9"
-tokio-core = "0.1.8"
 log = "0.3"
-
-[dev-dependencies]
-tokio-timer = "0.1.0"
 
 [badges]
 is-it-maintained-open-issues = { repository = "diwic/dbus-rs" }
--- a/dbus-tokio/src/adriver.rs
+++ b/dbus-tokio/src/adriver.rs
@@ -2,9 +2,11 @@ use mio::{self, unix, Ready};
 use mio::unix::UnixReady;
 use std::io;
 use dbus::{Connection, ConnMsgs, Watch, WatchEvent, Message, MessageType, Error as DBusError};
-use futures::{Async, Future, Stream, Poll};
-use futures::sync::{oneshot, mpsc};
-use tokio_core::reactor::{PollEvented, Handle as CoreHandle};
+use futures::{self, Async, Future, Stream, Poll};
+use futures::task::Context;
+use futures::executor::LocalExecutor;
+use futures_channel::{oneshot, mpsc};
+use tokio_reactor::{Handle, PollEvented};
 use std::rc::Rc;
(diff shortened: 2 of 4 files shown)


diwic:
Cool! I'll just keep this PR lingering here then to help other people discover it, and merge once 0.2 is actually released.

As for your question about need_read/need_write, I think they have been renamed to `clear_read_ready` and `clear_write_ready`? At least from a quick look [here](https://dtantsur.github.io/rust-openstack/tokio_reactor/struct.PollEvented.html).

o01eg:
Actually `tokio` not going to support `futures` 0.2 and they wait for 0.3.

Still they made some breaking changes and moves from deprecated `tokio-core` to `tokio`.

rubdos:
> Actually tokio not going to support futures 0.2 and they wait for 0.3.

The difference between 0.1 and 0.2 will probably be a quite bit larger than 0.2 to 0.3, so it still makes sense to keep thing hanging until Tokio has updated :-)

zeenix:
seems that tokio still depends on 0.1. Is that in anticipation of the async/wait in the stdlib?

diwic:
@zeenix That's a question for Tokio rather than DBus of course. Mean while I've started to work on a new [dbus-futures](https://github.com/diwic/dbus-rs/tree/master/dbus-futures) crate for Futures 0.3.

rubdos:
> Mean while I've started to work on a new [dbus-futures](https://github.com/diwic/dbus-rs/tree/master/dbus-futures) crate for Futures 0.3.

Shall we close this pr then?

diwic:
> Shall we close this pr then?

Good question. Let's keep it open until dbus-futures supports a tokio reactor (right now it only supports a thin_main_loop reactor), just in case there's something in here that I can use when I (or someone else) writes that code.

diwic:
> tokio-reactor = { version = "0.1.1", features = ["futures2"]}

Btw, does anybody if there's a tokio-reactor equivalent that works with Futures 0.3? Because the `futures2` feature seems to have been removed.

zeenix:
> Btw, does anybody if there's a tokio-reactor equivalent that works with Futures 0.3? Because the `futures2` feature seems to have been removed

I asked on IRC and was pointed to https://github.com/tokio-rs/tokio/blob/master/tokio-async-await/src/compat/backward.rs

diwic:
> I asked on IRC and was pointed to https://github.com/tokio-rs/tokio/blob/master/tokio-async-await/src/compat/backward.rs

Thanks. So I did an attempt to write tokio support for dbus-futures this evening. I can get it to build, but I can't get the tests to build, because that's where I need the compat layer, and so far every attempt to select some version for either futures (has a feature "compat") or tokio (has a feature "async-await-preview") has caused the dependency resolver to fail. Anyone wanna give it a try? I just want to get past the dependency resolver, then I can probably build & test further myself, as it probably isn't working yet.

SirWindfield:
Wasn't `futures` added to `stdlib` some days ago?

diwic:
Things have moved on, with new ways to handle things. This is probably very obsolete by now.

PR #140, opened 2018-08-01
Migrate from tokio-core to tokio.

### LINKED ISSUE #138: Migrate from tokio-core to tokio
Crate `tokio-core` is considered deprecated now.

o01eg:
Fixes #138.

It uses `current_thread::Runtime` because `ADriver` is `!Send`.

### DIFF
--- a/dbus-tokio/Cargo.toml
+++ b/dbus-tokio/Cargo.toml
@@ -14,11 +14,11 @@ categories = ["os::unix-apis", "api-bindings", "asynchronous"]
 dbus = { path = "../dbus", version = "0.6" }
 futures = "0.1.12"
 mio = "0.6.9"
-tokio-core = "0.1.8"
+tokio = "0.1.7"
 log = "0.3"
 
 [dev-dependencies]
-tokio-timer = "0.1.0"
+tokio-timer = "0.2.4"
 
 [badges]
 is-it-maintained-open-issues = { repository = "diwic/dbus-rs" }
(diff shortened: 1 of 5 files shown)


o01eg:
Looks like examples have issues from previous version.

diwic:
Thanks for the patch!

(Sorry for the slow reply - I've been on summer vacation.)

This is a large patch dealing with things I don't really know enough about, so code wise, I guess I could merge it - assuming you have tested all examples and found them to work.

But it also seems to combine with whitespace changes; in particular lots of more line breaks. Some are okay, but most of the shorter  `if a { b } else { c }` is better kept on one line (see [Single line if else](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/expressions.md)).

o01eg:
Those files was simply processed by `rustfmt` from nightly.

diwic:
Okay, so do you mind making a version of your patch which contains the code changes only, i e without the changes made by `rustfmt`?

o01eg:
I've made commit without style changes.

diwic:
Thanks a lot!

You seem to be right in that the examples don't work correctly...I'll try to look into that.

diwic:
Okay, the examples are working now.

marcelbuesing:
Great work! @diwic could you release a new version of the crate that contains the changes ?

diwic:
@marcelbuesing done now, enjoy!
Not sure if this was a breaking change or not but I did a 0.2.1 -> 0.3.0 version jump just in case.

PR #147, opened 2018-09-17
dbus: Implement reborrow for cstring wrappers

jvff:
This is my attempt at allowing references to `Path`s to be passed to `Connection::with_path`. I believe that with this solution it is similar to doing the following, but with safe code:

```
let borrowed_path = unsafe { Path::from_slice_unchecked(original_path.as_bytes()) };
```

### DIFF
--- a/dbus/src/strings.rs
+++ b/dbus/src/strings.rs
@@ -2,7 +2,7 @@
 
 use std::{str, fmt, ops, default, hash};
 use std::ffi::{CStr, CString};
-use std::borrow::Cow;
+use std::borrow::{Borrow, Cow};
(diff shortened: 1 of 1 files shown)


diwic:
Thanks for the patch! I think it looks correct (a test case would make it even better, but I'm okay with merging as it is too). 

Do you have any hesitations about it yourself, as you say that this is your "attempt" only?

jvff:
No hesitations, I called it an attempt because I think there might be other approaches as well :)

I wrote a fairly simple test case. Let me know if you have any comments on how to improve it.

diwic:
Thanks a lot!

PR #153, opened 2018-10-13
codegen: allow trivial_casts

hoodie:


diwic:
Thanks for the PR! However, as per answer in #154 I'm closing it in favor of changing the generated code instead.

PR #159, opened 2019-01-03
Rework appending of value types in iter_append()

### LINKED ISSUE #158: u16 values not properly marshalled on s390x
On s390x, a big-endian 64 bit architecture, we are seeing nonzero u16 property values being exported as 0, whereas u64 property values are correct.

In looking at the code, `iter_append_basic` in message.rs looks questionable to me:
```
unsafe {
    let p: *const c_void = mem::transmute(&v);
    ffi::dbus_message_iter_append_basic(i, t as c_int, p);
}
```
I think this may cause the problem, because p is a pointer to the entire i64 value. If libdbus expects p to point at a value with size based on t (the type) then it will get the upper zeroed bits rather than the lower significant bits.

@tasleson and I are still working to characterize this behavior a little more, but wanted to open an issue as a heads up, and if you had any thoughts on if you think this spot might be to blame, or another area perhaps. Issues like this are a pain for everyone to detect and fix because big endian is so scarce these days.

agrover:
iter_append_basic() doesn't quite work right on big-endian machines. It
results in the high bits of the i64 being parsed instead of the lower
bits. There's no elegant way to fix this (that I can see) due to the
differing widths of values, so open-code each value type instead of using
a helper function.

Call values v instead of b.

Instead of using mem::transmute(), use two casts: one to *const ptr of
<type>, and then one from that to *const c_void to match ffi function
signature.

Also do UnixFd and Double in this manner, allowing iter_append_basic() and
iter_append_f64() helper functions to be removed.

This also may fix an overflow issue in converting unsigned values to
their signed counterparts.

fixes #158

### DIFF
--- a/dbus/src/message.rs
+++ b/dbus/src/message.rs
@@ -400,10 +400,10 @@ impl MessageItem {
         v
     }
 
-    fn iter_append_basic(&self, i: &mut ffi::DBusMessageIter, v: i64) {
+    fn iter_append_basic<T>(&self, i: &mut ffi::DBusMessageIter, v: T) {
         let t = self.array_type();
+        let p = &v as *const _ as *const c_void;
         unsafe {
-            let p: *const c_void = mem::transmute(&v);
             ffi::dbus_message_iter_append_basic(i, t as c_int, p);
         }
     }
@@ -415,15 +415,15 @@ impl MessageItem {
                 let p = mem::transmute(&c);
                 ffi::dbus_message_iter_append_basic(i, ffi::DBUS_TYPE_STRING, p);
             },
-            &MessageItem::Bool(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::Byte(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::Int16(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::Int32(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::Int64(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::UInt16(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::UInt32(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::UInt64(b) => self.iter_append_basic(i, b as i64),
-            &MessageItem::UnixFd(ref b) => self.iter_append_basic(i, b.as_raw_fd() as i64),
+            &MessageItem::Bool(b) => self.iter_append_basic(i, b),
+            &MessageItem::Byte(b) => self.iter_append_basic(i, b),
+            &MessageItem::Int16(b) => self.iter_append_basic(i, b),
+            &MessageItem::Int32(b) => self.iter_append_basic(i, b),
+            &MessageItem::Int64(b) => self.iter_append_basic(i, b),
+            &MessageItem::UInt16(b) => self.iter_append_basic(i, b),
+            &MessageItem::UInt32(b) => self.iter_append_basic(i, b),
+            &MessageItem::UInt64(b) => self.iter_append_basic(i, b),
+            &MessageItem::UnixFd(ref b) => self.iter_append_basic(i, b.as_raw_fd()),
             &MessageItem::Double(b) => iter_append_f64(i, b),
             &MessageItem::Array(ref a) => iter_append_array(i, &a.v, a.element_signature()),
             &MessageItem::Struct(ref v) => iter_append_struct(i, &**v),


agrover:
@tasleson can you verify this resolves the issue on s390x?

diwic:
Still using MessageItem, huh?

Anyway, thanks for looking into it! 

As for some code review, repetition is not the end of the world but I think you could instead make `iter_append_basic` generic, like:

    fn iter_append_basic<T>(&self, i: &mut ffi::DBusMessageIter, v: T) {
        let t = self.array_type() as c_int;
        let p = &v as *const T as *const c_void;
        unsafe {
            ffi::dbus_message_iter_append_basic(i, t, p);
        }
    }

If that solves the problem as well, I believe that would be more elegant, right? That function should work for f64 too.

Also is fd:s guaranteed to be i32 on all platforms?

tasleson:
@agrover @diwic I tried this out and it doesn't appear to address the issue.  I then put in a `println!` to ensure we were going through this modified code path to only find out that we are not.  Further analysis shows that we are going through the code path of src/arg/basic_impl function `arg_append_basic`.  I'm trying to see how we can incorporate a change here that might alleviate the issue.

``` rust
fn arg_append_basic(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: i64) {
    let p = &v as *const _ as *const c_void;
    unsafe {
        check("dbus_message_iter_append_basic", ffi::dbus_message_iter_append_basic(i, arg_type as c_int, p));
    };
}
```

tasleson:
``` diff
 fn arg_append_basic(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: i64) {
-    let p = &v as *const _ as *const c_void;
+    // let p = &v as *const _ as *const c_void;
+ 
+    println!("arg_append_basic {:?} : {}", arg_type, v);
+
+    let p = match arg_type {
+        ArgType::UInt16 =>  {
+            let l = v as u16;
+            &l as *const _ as *const c_void
+        },
+        _ => &v as *const _ as *const c_void,
+    };
+
     unsafe {
         check("dbus_message_iter_append_basic", ffi::dbus_message_iter_append_basic(i, arg_type as c_int, p));
     };
```
This appears to work and addresses this specific issue, would need to expand for other types.  Is there a better way to approach this?

diwic:
@tasleson So it looks like we should make `arg_append_basic` generic as well as `iter_append_basic`, right? Like this:

    fn arg_append_basic<T>(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: T) {
        let p = &v as *const _ as *const c_void;
        unsafe {
            check("dbus_message_iter_append_basic", ffi::dbus_message_iter_append_basic(i, arg_type as c_int, p));
        };
    }

tasleson:
@diwic I just tried your proposed implementatioand.   It compiles fine, but does not solve the issue.  Not sure why at the moment.

tasleson:
@diwic

@agrover  found some other changes that needed to be made.  With the following things appear to work, but obviously this isn't any type of exhaustive test :-)

``` diff
-fn arg_append_basic(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: i64) {
+
+fn arg_append_basic<T>(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: T) {
     let p = &v as *const _ as *const c_void;
     unsafe {
         check("dbus_message_iter_append_basic", ffi::dbus_message_iter_append_basic(i, arg_type as c_int, p));
@@ -68,7 +69,7 @@ impl Arg for $t {
 }
 
 impl Append for $t {
-    fn append(self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, self as i64) }
+    fn append(self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, self) }
 }
 
 impl<'a> Get<'a> for $t {
@@ -81,7 +82,7 @@ impl RefArg for $t {
     #[inline]
     fn signature(&self) -> Signature<'static> { unsafe { Signature::from_slice_unchecked($f) } }
     #[inline]
-    fn append(&self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, *self as i64) }
+    fn append(&self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, *self) }
     #[inline]
     fn as_any(&self) -> &any::Any { self }
     #[inline]
@@ -240,7 +241,7 @@ impl Arg for OwnedFd {
 impl Append for OwnedFd -fn arg_append_basic(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: i64) {
+
+fn arg_append_basic<T>(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: T) {
     let p = &v as *const _ as *const c_void;
     unsafe {
         check("dbus_message_iter_append_basic", ffi::dbus_message_iter_append_basic(i, arg_type as c_int, p));
@@ -68,7 +69,7 @@ impl Arg for $t {
 }
 
 impl Append for $t {
-    fn append(self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, self as i64) }
+    fn append(self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, self) }
 }
 
 impl<'a> Get<'a> for $t {
@@ -81,7 +82,7 @@ impl RefArg for $t {
     #[inline]
     fn signature(&self) -> Signature<'static> { unsafe { Signature::from_slice_unchecked($f) } }
     #[inline]
-    fn append(&self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, *self as i64) }
+    fn append(&self, i: &mut IterAppend) { arg_append_basic(&mut i.0, ArgType::$s, *self) }
     #[inline]
     fn as_any(&self) -> &any::Any { self }
     #[inline]
@@ -240,7 +241,7 @@ impl Arg for OwnedFd {
 impl Append for OwnedFd {
     fn append(self, i: &mut IterAppend) {
         use std::os::unix::io::AsRawFd;
-        arg_append_basic(&mut i.0, ArgType::UnixFd, self.as_raw_fd() as i64)
+        arg_append_basic(&mut i.0, ArgType::UnixFd, self.as_raw_fd())
{
     fn append(self, i: &mut IterAppend) {
         use std::os::unix::io::AsRawFd;
-        arg_append_basic(&mut i.0, ArgType::UnixFd, self.as_raw_fd() as i64)
+        arg_append_basic(&mut i.0, ArgType::UnixFd, self.as_raw_fd())

```

agrover:
While Tony works on testing his Arg fix, I've pushed an updated MessageItem endian fix. I'm assuming also fixing MessageItem is still worthwhile :)

tasleson:
@diwic @agrover 

I created https://github.com/tasleson/misc-tools/tree/master/dbus_values to validate.  I added a method which simply returns the values passed in and we fail on big endian, pass on little endian.  Guessing `arg_get_basic` needs correction too.  Getting values back from the service seems to work with my diff I posted above.

tasleson:
See https://github.com/diwic/dbus-rs/pull/160

PR #160, opened 2019-01-04
Big endian args fix

tasleson:
These changes were tested with: https://github.com/tasleson/misc-tools/tree/master/dbus_values on x86_64 and s390x.  The changes for `arg_get_basic` is a non-elegant solution.  Maybe someone else can find a better way of handling it.  I know of no way to have a return type be generic except for returning an enum, but that still wouldn't change how `arg_get_basic`is implemented internally.

### DIFF
--- a/dbus/src/arg/basic_impl.rs
+++ b/dbus/src/arg/basic_impl.rs
@@ -16,10 +16,54 @@ fn arg_append_basic<T>(i: *mut ffi::DBusMessageIter, arg_type: ArgType, v: T) {
 
 fn arg_get_basic(i: *mut ffi::DBusMessageIter, arg_type: ArgType) -> Option<i64> {
     let mut c = 0i64;
+
     unsafe {
         if ffi::dbus_message_iter_get_arg_type(i) != arg_type as c_int { return None };
-        ffi::dbus_message_iter_get_basic(i, &mut c as *mut _ as *mut c_void);
     }
+
+    match arg_type {
+        ArgType::Byte => {
+        let mut t = 0u8;
+            unsafe {
+                ffi::dbus_message_iter_get_basic(i, &mut t as *mut _ as *mut c_void);
+            }
+            c = t as i64;
+        },
+        ArgType::Int16 => {
+            let mut t = 0i16;
+            unsafe {
+                ffi::dbus_message_iter_get_basic(i, &mut t as *mut _ as *mut c_void);
+            }
+            c = t as i64;
+        },
+        ArgType::UInt16 => {
+            let mut t = 0u16;
+            unsafe {
+                ffi::dbus_message_iter_get_basic(i, &mut t as *mut _ as *mut c_void);
+            }
+            c = t as i64;
+        },
+        ArgType::Int32 => {
+            let mut t = 0i32;
+            unsafe {
+                ffi::dbus_message_iter_get_basic(i, &mut t as *mut _ as *mut c_void);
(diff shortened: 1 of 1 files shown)


diwic:
Hi! I cherrypicked the arg_append_basic one and wrote my own for arg_get_basic. Can you check that  this is resolved in git master now?

tasleson:
> Hi! I cherrypicked the arg_append_basic one and wrote my own for arg_get_basic. Can you check that this is resolved in git master now?

@diwic @agrover Changes committed to master do not resolve the issue with `arg_get_basic`.

diwic:
@tasleson Then I'm afraid you have to look deeper yourself and explain to me why my new version of arg_get_basic does not work. I don't have access to a big endian machine so I can't debug it myself...

tasleson:
@diwic  @agrover The problem isn't with the changes in master with the corrections to `arg_get_basic`,  it's when the service appends values to send back to the client they are incorrect.  My local testing branch incorrectly had @agrover first correction for `iter_append_basic` with my correction to `arg_get_basic`.  When I tested the latest upstream main, no corrections for it are present.  However, if I pick up the latest PR from @agrover https://github.com/diwic/dbus-rs/pull/159 and use it with current master I assert on x86_64 and Z.
```
dbus[21125]: arguments to dbus_message_iter_append_basic() were incorrect, assertion "*bool_p == 0 || *bool_p == 1" failed in file ../../dbus/dbus-message.c line 2777.
This is normally a bug in some application using the D-Bus library.
```

We need @agrover changes too, but the current solution needs some corrections.  I haven't tried taking master and applying @agrover  initial correction.

tasleson:
@diwic @agrover OK, with the following change I pass my tests on Z and don't assert.

``` diff
diff --git a/dbus/src/message.rs b/dbus/src/message.rs
index cad61f8..84822fb 100644
--- a/dbus/src/message.rs
+++ b/dbus/src/message.rs
@@ -415,7 +415,7 @@ impl MessageItem {
                 let p = mem::transmute(&c);
                 ffi::dbus_message_iter_append_basic(i, ffi::DBUS_TYPE_STRING, p);
             },
-            &MessageItem::Bool(b) => self.iter_append_basic(i, b),
+            &MessageItem::Bool(b) => self.iter_append_basic(i, if b {1} else {0}),
             &MessageItem::Byte(b) => self.iter_append_basic(i, b),
             &MessageItem::Int16(b) => self.iter_append_basic(i, b),
             &MessageItem::Int32(b) => self.iter_append_basic(i, b),
```

This change is stacked on current master with @agrover PR https://github.com/diwic/dbus-rs/pull/159

diwic:
@tasleson Ok, so in master we now have generic versions of iter_get_basic, iter_append_basic, arg_get_basic and arg_append_basic. I've included your fixup for bool.

Is all working the way it should now?

Also, you had a nice test program there, I wonder how feasible it is to include that test into dbus-rs?

tasleson:
@diwic @agrover 

> Is all working the way it should now?

Testing of current master passes my test, although I don't think we are exercising the `iter_get_basic` path with it.  I may extend the test to ensure all code paths are covered.  Also it would be good to tests for strings and container types.

> Also, you had a nice test program there, I wonder how feasible it is to include that test into dbus-rs?

Not sure, the client is written in python3.  I figured if we wrote all (client & service) in rust using the dbus library we could potentially hide an issue.

PR #163, opened 2019-01-17
generate codegen testcode at buildtime

### LINKED ISSUE #162: dbus-codegen: Generate code during build for tests
Currently the generated code being tested as part of `cargo test` is committed to git and hence the tests don't actually test the current dbus-codegen binary, which IMO beats the main point of the tests.

In any case, the XML the code is generated from, needs to be committed to git so devs can regenerate the test code to be able to easily test any changes to dbus-codegen.

zeenix:
Fixes #162


diwic:
Really nice, thanks for cleaning that up!

I was wondering if one could use the path attribute `#[path = "foo.rs"]` instead of `include!` in order to not have to remove the `allow(dead_code)` part? Or if that is not possible, write a file in OUT_DIR which can be included and just contains something like `mod policykit.rs; pub use policykit::*;`? 
Because removing allow(dead_code) is going to be quite annoying I'm afraid...

zeenix:
> Really nice, thanks for cleaning that up!

No worries. :)

> I was wondering if one could use the path attribute `#[path = "foo.rs"]`

I'll have a look at both approaches you mentioned but what is the deal with `allow(dead_code)`? I didn't see any warnings about deadcode in `cargo build/test` output.

zeenix:
Also, we can have the `allow(dead_code)` in the including module instead and advertise that as *the* way to use generated code. If you recall I was wondering how I'll use generated code in my own (external) crate and seems this is a good way to do things and seems better than committing generated code in VCS.

diwic:
> I'll have a look at both approaches you mentioned but what is the deal with allow(dead_code)? I didn't see any warnings about deadcode in cargo build/test output.

dbus-codegen, when run against some server, that server usually returns many interfaces, including Properties and Introspectable. We generate code for them, but it is seldom used by the program that's including the module. They end up being dead code. 

 > Also, we can have the allow(dead_code) in the including module instead and advertise that as the way to use generated code.

We could, I would consider this the fallback option if neither of my proposed solutions above work.

zeenix:
> dbus-codegen, when run against some server, that server usually returns many interfaces, including Properties and Introspectable. We generate code for them, but it is seldom used by the program that's including the module. They end up being dead code.

Yeah i understand that bit but as I said rustc didn't complain at all about dead code with my first patch only (i-e without my tests architecture changes). The reason is that there is already an `#![allow(dead_code)]` line above the `mod generated` line in the `test1.rs` and `test2.rs`. So I don't see the point of finding an alternative.

> We could, I would consider this the fallback option if neither of my proposed solutions above work.

I looked into the `path` option you mentioned and the problem with that is that it expects the file in the same source directory if I understood the docs correctly: "the file path is relative to the directory the source file is located" https://doc.rust-lang.org/reference/items/modules.html

The second alternative you mentioned, sounds fine to me and I'll have a look at that, if I didn't convince you above. :)

diwic:
Ok, so after some more code review I realized it was a bit silly that we just generated the asref file without doing anything with it so I added a test for that as well. Apparently the "Generic" part isn't working yet...

zeenix:
> Ok, so after some more code review I realized it was a bit silly that we just generated the asref file without doing anything with it so I added a test for that as well. Apparently the "Generic" part isn't working yet...

Cool, things you find out when you add tests :)

PR #165, opened 2019-01-30
dbus-tokio: return a plain Future on method call

lucab:
This hides the internal details of `method_call` helper struct, returning
a plain `impl Future` instead and removing the need for the additional
`Result` wrapper.

### DIFF
--- a/dbus-tokio/examples/tokio_client.rs
+++ b/dbus-tokio/examples/tokio_client.rs
@@ -46,7 +46,7 @@ fn main() {
         //TODO: try to handle error when calling on "/"
         let m = Message::new_method_call("com.example.dbustest", "/hello", "com.example.dbustest", "Hello")
             .unwrap().append1(500u32);
-        aconn.method_call(m).unwrap().then(|reply| {
+        aconn.method_call(m).then(|reply| {
             let m = reply.unwrap();
             let msg: &str = m.get1().unwrap();
             println!("{}", msg);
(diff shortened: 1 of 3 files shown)


lucab:
I understand (even if so far I haven't had a case where I wanted to name the type of a future, only the opposite), happy to rework that way. One of my goals though was to get rid of the outer `Result` and just use the future error. Do you think there is a way to do that while keeping the struct?

Also, I'd like to cross-link https://github.com/seanmonstar/reqwest/issues/205 here which is a similar discussion going in the opposite direction.

lucab:
I split the TODO-fix in # because I still haven't found a satisfying solution for the return type.

> Having looked at the code for a minute I would be okay with this change, i e, to return an AMethodCall rather than a Result<AMethodCall, &'static str>. I guess this means moving most of the code from fn method_call to AMethodCall::poll, unless you have a better idea.

I've experimented a bit with this, but I'm not very happy with what I sketched. `poll` is an hot-path and having an "unlikely initial error" check in there just feels bad.

Another idea that came to mind is aliasing the result type as `pub type CallResult = Box<Future<I=Message, E=DBusError>` which is similar to the initial `impl Future` but gives the consumer a proper named type (trade-off: boxed allocation). That allows a single level of future-error-handling without embedding those checks in `poll`. What do you think of that?

diwic:
Hmm, I'm not sure about this, but would not returning a `Box<Future>` result in an extra virtual dispatch, which would be worse (performance wise) than an "unlikely initial error" check?

Also, I'm not sure the existing solution is so bad, could you explain what's wrong about it? For me it makes sense to return errors ASAP.

diwic:
I'm closing this due to inactivity. Feel free to reopen if you want to discuss further.

PR #169, opened 2019-02-12
replace x86_64 specific bits in rtkit example with calls to libc

jcristau:


diwic:
Thanks!

Although this example is very old and maybe should be dropped or rewritten altogether.

PR #177, opened 2019-04-13
Update dbus-tokio/example/tokio_server.rs:

LinkTed:
Use `Handle::default()` instead of `Handle::current()`, because it is deprecated.


diwic:
Thanks!

Issue #173, opened 2019-04-07
dbus-codegen: Option to specify input files as args


zeenix:
I'd expect this as a user anyway but without this option I had to write a custom python script to call this from a meson `custom_target()`.

diwic:
Sounds like an easy task, feel free to submit a PR.

PR #180, opened 2019-04-24
dbus-codegen: Option to specify input files as args

inzamam-iqbal:
Users can call dbus-codegen-rust [path_to_file] to specify the
path to D-Bus XML Introspection data file.

fixes: #173

### DIFF
--- a/dbus-codegen/src/main.rs
+++ b/dbus-codegen/src/main.rs
@@ -47,8 +47,13 @@ Defaults to 'RefClosure'."))
              .help("If present, skips a specific prefix for interface names, e g 'org.freedesktop.DBus.'."))
         .arg(clap::Arg::with_name("futures").short("f").long("futures")
              .help("Generates code to use with futures 0.3 (experimental)"))
+        .arg(clap::Arg::with_name("file").required(false).help("D-Bus XML Introspection file"))
         .get_matches();
 
+    if matches.is_present("destination") && matches.is_present("file") {
+        panic!("Expected either xml file path as argument or destination option. But both are provided.");
+    }
+	
     let s = 
     if let Some(dest) = matches.value_of("destination") {
         let path = matches.value_of("path").unwrap_or("/");
@@ -56,6 +61,8 @@ Defaults to 'RefClosure'."))
         let c = dbus::Connection::get_private(bus).unwrap();
         let p = c.with_path(dest, path, 10000);
         p.introspect().unwrap()
+    } else if let Some(file_path) = matches.value_of("file")  {
+        std::fs::read_to_string(file_path.to_string()).unwrap()
     } else {
         let mut s = String::new();
         (&mut std::io::stdin() as &mut std::io::Read).read_to_string(&mut s).unwrap();


dbus-codegen/src/main.rs
```diff
@@ -47,6 +47,8 @@ Defaults to 'RefClosure'."))
              .help("If present, skips a specific prefix for interface names, e g 'org.freedesktop.DBus.'."))
         .arg(clap::Arg::with_name("futures").short("f").long("futures")
              .help("Generates code to use with futures 0.3 (experimental)"))
+        .arg(clap::Arg::with_name("file").long("file").takes_value(true).value_name("FILE")
```
zeenix: I think we want an argument, not an option here. i-e `dbus-codegen-rust [OPTIONS] [INTERFACE_XML_FILE]`. See [here](https://github.com/zeenix/gps-share/blob/master/src/cmdline_config.rs#L32) for an example how that's achieved (the device arg there is an argument, not an option).

dbus-codegen/src/main.rs
```diff
@@ -56,6 +58,10 @@ Defaults to 'RefClosure'."))
         let c = dbus::Connection::get_private(bus).unwrap();
         let p = c.with_path(dest, path, 10000);
         p.introspect().unwrap()
+    } else if let Some(_file_path) = matches.value_of("file")  {
```
zeenix: Why ignore the path with `_` prefix if you will use it in the block?

dbus-codegen/src/main.rs
```diff
@@ -56,6 +58,10 @@ Defaults to 'RefClosure'."))
...
+    	let mut s = String::new();
+    	s=std::fs::read_to_string(matches.value_of("file").unwrap().to_string()).unwrap();
```
zeenix: * this unwrap can be avoided if you just use the value you unpack above if the if condition.
* Some space before and after the `=` would be nice.

dbus-codegen/src/main.rs
```diff
@@ -56,6 +58,10 @@ Defaults to 'RefClosure'."))
         let c = dbus::Connection::get_private(bus).unwrap();
         let p = c.with_path(dest, path, 10000);
         p.introspect().unwrap()
+    } else if let Some(_file_path) = matches.value_of("file")  {
+    	let mut s = String::new();
```
zeenix: This is quite redundant. Just add a `let` before the next statement to create and initialize the variable where it's needed. This is one thing that's different from C that you'll want to get used to.

dbus-codegen/src/main.rs
zeenix: I think we should error out if both `destination` and `file` arguments are provided as otherwise it's ambiguous, which will be used and we really don't need to support the case of both being provided. @diwic wouldn't you agree?

zeenix:
It's not a bad commit message at all but as requested before, kindly follow the English grammar rules in the description part:

* Capitalize the first letter of a every sentence/phrase.
* A '.' to indicate end of a sentence/phrase.

I hope I won't need to point these out again. :)

dbus-codegen/src/main.rs
diwic: I guess I prefer erroring out, but it's not a big deal.

dbus-codegen/src/main.rs
diwic: I kind of prefer keeping it as an option, unless there is more to it than just a matter of taste?

dbus-codegen/src/main.rs
diwic: Actually you don't need `s` at all.

diwic:
So, I suppose that @inzamam-iqbal is someone @zeenix mentors, perhaps a GSOC student? Anyhow, a warm welcome if you are new to open source collaboration! Review conversations like the above can have a quite dry/boring language, but don't worry about that.

When I review pull requests I try to see to the whole picture and to avoid pulling buggy code, but not be very fuzzed over details. Sometimes easier said than done though. 🙂

dbus-codegen/src/main.rs
zeenix: Two reasons for arg:

1. I think that's the most typical way to provide input to DBus code generation tool.
2. It's consistent with existing tools. See gdbus-codegen and qdbusxml2cpp for example.

zeenix:
@diwic @inzamam-iqbal is a potential GSoC student indeed and if they get chosen, I'll be mentoring indeed.

dbus-codegen/src/main.rs
inzamam-iqbal: @zeenix @diwic The existing code looks for input whenever `destination` is not provided. Do I have too still keep that code such that it executes if both `destination` and `file` are not provided?

diwic: @inzamam-iqbal The code should behave like this:

 * Both destination and file - panic with relevant error message
 * destination only - read from destination (like today)
 * file only - read from file
 * neither file, nor destination - read from stdin (like today)

Does that answer your question?

inzamam-iqbal: @diwic yes I have updated the PR like that

inzamam-iqbal:
@zeenix @diwic can you please review it now?

dbus-codegen/src/main.rs
```diff
@@ -47,15 +47,23 @@ Defaults to 'RefClosure'."))
...
+	
     let s = 
     if let Some(dest) = matches.value_of("destination") {
         let path = matches.value_of("path").unwrap_or("/");
         let bus = if matches.is_present("systembus") { dbus::BusType::System } else { dbus::BusType::Session };
         let c = dbus::Connection::get_private(bus).unwrap();
         let p = c.with_path(dest, path, 10000);
         p.introspect().unwrap()
+    } else if let Some(file_path) = matches.value_of("file")  {
+    	let s = std::fs::read_to_string(file_path.to_string()).unwrap();
```
diwic:
```suggestion
    	std::fs::read_to_string(file_path.to_string()).unwrap()
```

dbus-codegen/src/main.rs
```diff
@@ -47,15 +47,23 @@ Defaults to 'RefClosure'."))
...
+    	s
```
diwic:
```suggestion
    	
```

dbus-codegen/src/main.rs
```diff
@@ -47,15 +47,23 @@ Defaults to 'RefClosure'."))
              .help("If present, skips a specific prefix for interface names, e g 'org.freedesktop.DBus.'."))
         .arg(clap::Arg::with_name("futures").short("f").long("futures")
              .help("Generates code to use with futures 0.3 (experimental)"))
+        .arg(clap::Arg::with_name("file").required(false).help("D-Bus XML Introspection file"))
         .get_matches();
 
+	if matches.value_of("destination").is_some() && matches.value_of("file").is_some() {
```
diwic: Wrong indentation

diwic:
@zeenix you can just test it and see that it works the way you expect. If it does I'll merge with the suggested changes above.

dbus-codegen/src/main.rs
diwic: To clarify: Please use four spaces per indentation level (not tabs).

dbus-codegen/src/main.rs
inzamam-iqbal: This doesn't work. A build error is thrown. 
Screenshot from 2019-04-26 01-18-41

inzamam-iqbal:
@zeenix @diwic made the suggested changes

dbus-codegen/src/main.rs
```diff
@@ -47,15 +47,22 @@ Defaults to 'RefClosure'."))
...
     let s = 
     if let Some(dest) = matches.value_of("destination") {
         let path = matches.value_of("path").unwrap_or("/");
         let bus = if matches.is_present("systembus") { dbus::BusType::System } else { dbus::BusType::Session };
         let c = dbus::Connection::get_private(bus).unwrap();
         let p = c.with_path(dest, path, 10000);
         p.introspect().unwrap()
+    } else if let Some(file_path) = matches.value_of("file")  {
+        std::fs::read_to_string(file_path.to_string()).unwrap()
     } else {
```
zeenix: I'm very sorry but i only saw this `else` block now, which takes the interface XML from stdin if neither `destination` nor `file` arguments are specified. So I'm afraid we need to remove the check above for either `destination` or `file` being specified.

zeenix: sorry, my bad.

dbus-codegen/src/main.rs
zeenix: we should setup `rustfmt` check in the CI (and CI itself) I think.

zeenix: Here is an example (although for gitlab): https://gitlab.gnome.org/GNOME/fractal/blob/master/.gitlab-ci.yml#L23

dbus-codegen/src/main.rs
inzamam-iqbal: @zeenix In the above check we are checking whether both destination and file are provided. If both are provided we panic. that's fine isn't it? because if both are provided it is ambiguous what to use?

zeenix: Ah yes, sorry I had a long day but good that you can point out when I'm wrong. We'll need that quality. :)

dbus-codegen/src/main.rs
```diff
@@ -47,15 +47,22 @@ Defaults to 'RefClosure'."))
              .help("If present, skips a specific prefix for interface names, e g 'org.freedesktop.DBus.'."))
         .arg(clap::Arg::with_name("futures").short("f").long("futures")
              .help("Generates code to use with futures 0.3 (experimental)"))
+        .arg(clap::Arg::with_name("file").required(false).help("D-Bus XML Introspection file"))
         .get_matches();
 
+    if matches.value_of("destination").is_some() && matches.value_of("file").is_some() {
```
zeenix: You should make use of https://kbknapp.github.io/clap-rs/clap/struct.ArgMatches.html#method.is_present here.

inzamam-iqbal: I'm new to rust. tank you for the link. I have updated the MR

zeenix: I'm not very experienced either and with each day, I forget more. :( But together I hope we can tame the beast. :) Thanks, I'll test it.

zeenix:
I tested the code and it works. Thanks. @diwic please merge this PR.

@inzamam-iqbal I realized that we will also need `--output FILE` parameter to tell the code generator to output to a file instead of stdout for being able to ditch the python script for meson build. Could you please create an issue and when you have time, please fix that as well? Thanks.

inzamam-iqbal:
@zeenix Sure I will

PR #183, opened 2019-05-09
futures: Don't assume local path to thin_main_loop crate

zeenix:
Not everyone will have `thin_main_loop` checked out and it will certainly fail when trying to install `dbus-codegen-rust` through `cargo install`:

```
$ cargo install --force --git https://github.com/diwic/dbus-rs.git  --bin dbus-codegen-rust --no-default-features
    Updating git repository `https://github.com/diwic/dbus-rs.git`
error: Could not find `Cargo.toml` in `/home/zeenix/.cargo/git/checkouts/dbus-rs-295c8703110e4ae3/thin_main_loop`
```

### DIFF
--- a/dbus-futures/Cargo.toml
+++ b/dbus-futures/Cargo.toml
@@ -9,8 +9,8 @@ futures-preview = { version = "0.3.0-alpha.12" }
 dbus = { path = "../dbus" }
 
 [dependencies.thin_main_loop]
-# git = "https://github.com/diwic/thin_main_loop.git"
-path = "../../thin_main_loop"
+git = "https://github.com/diwic/thin_main_loop.git"
+#path = "../../thin_main_loop"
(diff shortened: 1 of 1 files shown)


zeenix:
Turns out `cargo install` tries to build `dbus-futures` no matter what. :-( So this patch won't help with that and I'll have to have a fork with `dbus-futures` completely removed for it to work. This patch is still good though.

diwic:
This looks like a missing feature in cargo. I think the path and the git arguments can't be combined the way you are trying to do it.

As a workaround, do:

    git clone https://github.com/diwic/dbus-rs.git
    cargo install --path dbus-rs/dbus-codegen --force

zeenix:
> This looks like a missing feature in cargo.

Indeed. Thanks for the workaround. :)

PR #187, opened 2019-06-18
tokio_dbus: add compatibility with Actix

david-mcgillicuddy-moixa:
Hi, I (and my colleagues) are trying to use the async portion of this library (we're happily using the sync version) and the main issue we had was the requirement to hold a `current_thread::Runtime` so that the `ADriver` can be spawned which, as far as I can tell, Actix does not allow.

In this PR I have changed `AConnection::new` to use `current_thread::spawn` instead, allowing compatibility with any runtime that ultimately uses `current_thread::Runtime` but doesn't expose it directly. This unfortunately means that `AConnection::new` now returns a future since `current_thread::spawn` has to be called inside one.

Another possible option would be to make it accept a [TypedExecutor](https://docs.rs/tokio/0.1.21/tokio/executor/trait.TypedExecutor.html) which Actix's `AsyncContext` doesn't _currently_ implement but I don't think it'd be hard to do that separately.

Additionally I had issues with my `AConnection` being dropped while a `AMethodCall` was still active. This caused the future to never complete. To solve this I made `AMethodCall` keep an `Rc<AConnection>` to prevent the drop while it was still being used. I suspect that this means that the `oneshot::channel` mechanism isn't needed anymore since the future can just query the connection directly but didn't spent much time playing with alternatives.

Finally I ran `cargo fmt` more out of habit than anything else and bumped the version (sorry for the diff).

Happy to split this up, discuss alternatives etc. I've run the tests and the examples (together) and they're all successful.

In summary:
1. Remove `current_thread::Runtime` parameter from `AConnection::new`.
2. Prevent `AConnection` from being dropped while an `AMethodCall` exists.
3. cargo fmt & bump version

### DIFF
--- a/dbus-tokio/Cargo.toml
+++ b/dbus-tokio/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["David Henningsson <diwic@ubuntu.com>"]
 name = "dbus-tokio"
-version = "0.3.0"
+version = "0.4.0"
 
 description = "Makes it possible to use Tokio with D-Bus, which is a bus commonly used on Linux for inter-process communication."
 repository = "https://github.com/diwic/dbus-rs"
--- a/dbus-tokio/examples/tokio_client.rs
+++ b/dbus-tokio/examples/tokio_client.rs
@@ -7,19 +7,19 @@
 
 extern crate dbus;
 extern crate dbus_tokio;
+extern crate futures;
 extern crate tokio;
 extern crate tokio_timer;
-extern crate futures;
 
 use dbus::*;
 
-use std::rc::Rc;
-use tokio::reactor::Handle;
-use tokio::runtime::current_thread::Runtime;
-use tokio_timer::{clock, Interval};
-use std::time::Duration;
-use futures::{Stream, Future};
 use dbus_tokio::AConnection;
+use futures::{Future, Stream};
+use std::rc::Rc;
+use std::time::Duration;
+use tokio::reactor::Handle;
+use tokio::runtime::current_thread;
+use tokio_timer::{clock, Interval};
 
 fn main() {
     // Let's start by starting up a connection to the session bus. We do not register a name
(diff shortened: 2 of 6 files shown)


diwic:
Hi and thanks for the PR!

So just a few things first:

 * dbus-tokio is scheduled to be rewritten from scratch when Futures 0.3 becomes stable, which is probably just a few weeks from now.
 * `cargo fmt` usually makes things uglier, please avoid it by default
 
As for the patch, I've never used Actix so I'm not sure why it would not allow using `current_thread::Runtime`, could you elaborate on this part? Maybe would be good to have two constructors for AConnection, one which uses `current_thread::Runtime` and one which uses `current_thread::spawn`?

david-mcgillicuddy-moixa:
Yes I did somewhat regret it immediately after running!

Actix's behaviour (not that I had anything to do with it), is that while the _backing_ runtime is just `current_thread`, the [`AsyncContext.spawn` method](https://docs.rs/actix/0.8.3/actix/trait.AsyncContext.html#tymethod.spawn) exposed to users is per-actor. If it handed out the actual `current_thread::Runtime` then it couldn't enforce that.

The more I think about it, the more I feel like the `TypedExecutor` solution is better - it's "just" a generalisation of the existing solution than the major change that I had proposed.

As for 0.3 I suppose since I'm happy enough with the sync code I don't have any real need for this PR, but if it's possible to use a similar trait to allow for actix compatibility that would be great.

Feel free to close this PR and I can open a separate issue about the connection closing thing.

diwic:
If we take a step back; we need *something* that reacts to the D-Bus fd being ready for reading, reads and processes incoming messages somehow. And I believe, but I might be wrong because I don't know this as well as I probably should, that the *something* is a separate task. 

Then that task needs to be spawned on some executor somehow. Maybe the best way is to actually leave that to the user to decide? By returning both the `AConnection` and the `ADriver`, with instructions that the `ADriver` needs to be separately spawned? For Actix, `ADriver` would be spawned in some way, whereas in other environments, it would be spawned some other way. Does that make sense?

david-mcgillicuddy-moixa:
Yes that makes sense. All the other stuff we've been discussing is just the way to spawn that task.

I spent a while looking at the design of `tokio-serial` etc, and they use [tokio::reactor::PollEvented](https://docs.rs/tokio/0.1.21/tokio/reactor/struct.PollEvented2.html) which in turn uses [tokio::reactor::Registration](https://docs.rs/tokio/0.1.21/tokio/reactor/struct.Registration.html) to run their backing tasks instead of explicitly passing them to the user to run.

diwic:
So what you're saying is that `AConnection::new` should take a `Registration`? 

I think Registration is newer than dbus-tokio, which is why I didn't consider it while writing dbus-tokio years ago. But now that it is there, maybe it would make the most sense to use it. Would that be Actix compatible?

diwic:
There is something to keep in mind though. In rare cases, the number of fds can change during runtime. While we don't handle that now, I would like to be able to handle that case in the future. So the API needs to take that into account as well.

david-mcgillicuddy-moixa:
After some more reading, I see that you are already using `PollEvented` to watch individual file handles and then `ADriver` holds all those references and when it is polled it polls each in turn. 

I am not saying that `AConnection::new` should take a registration, but (and I'm really not sure about this) if something like `mio::Evented` was implemented for `ADriver` so that it woke and e.g. read in messages when any of its FDs woke, then `AConnection` would hold a reference to the underlying `ADriver` and then each future e.g. `AMethodCall` would hold a `Registration<ADriver>`.

As far as I can tell from the docs, `Registration` has a singleton mechanism:
> A registration represents an I/O resource registered with a Reactor such that it will receive task notifications on readiness. This is the lowest level API for integrating with a reactor.
The association between an I/O resource is made by calling register. Once the association is established, it remains established until the registration instance is dropped. Subsequent calls to register are no-ops.
And so polling any individual future e.g. a specific message or the messages stream would cause the dbus driver itself to check for readiness and potentially read data in and there's no need for the driver to be a separate task.

In short I guess what I'm proposing is that instead of having `ADriver` be a separate middleware that runs independently of its file handles, that 1) a reference to it exists inside the futures that get polled (i.e. in `AConnection` and then passed to any future that is spawned) and 2) whenever its FDs wake.
This would also get rid of the channel notification mechanism - polling an `AMethodCall` would involve it polling the `ADriver` and then checking for relevant replies.

diwic:
> In short I guess what I'm proposing is that instead of having `ADriver` be a separate middleware that runs independently of its file handles, that 1) a reference to it exists inside the futures that get polled (i.e. in `AConnection` and then passed to any future that is spawned) and 2) whenever its FDs wake.
> This would also get rid of the channel notification mechanism - polling an `AMethodCall` would involve it polling the `ADriver` and then checking for relevant replies.

This sounds indeed simpler, and simple is good 🙂 I'm trying to think this through to see if it is a reasonable approach or not. Here are some complications:

 * Connections are idle for the most part. I e, there is an active connection / unix socket to D-Bus, but there are no method calls currently in progress, so there are no corresponding futures either. What happens in this case, if a `org.freedesktop.DBus.Peer.Ping` comes in? If the FDs are not being polled at all (because there are no futures?) - what will make us (or libdbus library) to reply correctly?

 * Currently `AConnection` is not `Send`, I would like it to be for the rewrite. Even `Sync`, if possible. Polling a future would then require a mutex lock, but perhaps this is not a problem. But you should be able to make method calls in parallel from different threads.

david-mcgillicuddy-moixa:
- If some running is required just from a connection existing even when there are no method calls, then in this design that would mean that creating a connection (or perhaps not exactly a connection but a driver) returns a future, which when polled does the things that it needs to do. I think that this makes sense, because creating a connection is causing work to be done. We're basically back full circle to this PR's solution!
- Do you know how other languages' libraries handle multiple concurrent connections to a single "driver"? Is dbus inherently sequential i.e. there is always a single object that interacts with dbus itself or do they have independent "actors" in a client?

diwic:
> Do you know how other languages' libraries handle multiple concurrent connections to a single "driver"? Is dbus inherently sequential i.e. there is always a single object that interacts with dbus itself or do they have independent "actors" in a client?

Not really, I'm afraid. I think it depends a lot on whether or not that language is single- or multithreaded. The fact that Rust tries to support both makes matters more complex. The reactor/executor pattern that Tokio uses is a bit bewildering; but in essence I think it means that you should be able to have one task registering with the reactor (the driver), and then other tasks submitted to an executor, potentially in a thread pool, which in turn somehow uses that driver task...

david-mcgillicuddy-moixa:
Good discussion, I'm going to close this PR but let's pick this up again after the async ecosystem is more stable and work on async dbus can start!

PR #190, opened 2019-07-10
Default interface example

david-mcgillicuddy-moixa:
Hi, as requested I added a test for `default_interface` and also added it to the `adv_server` example - which I was then able to call from python.

The test for `default_interface` isn't _massively_ useful if you have a suggestion for a better one I'd be happy to set that up. As far as I can see the client can't make calls with no interface so we can't test it with the client part of this library.

I also had an issue when compiling the tests:
```
error[E0308]: mismatched types
   --> dbus/src/watch.rs:222:59
    |
222 |             assert!(unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, 1000) } > 0);
    |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u32, found u64

```
Which I think was due to an assumption about pointer size, which I fixed by instead casting to `libc::nfds_t` instead - the type that `libc::poll` is actually expecting.

Is it possible to release a new version including even just this change? I'd like to move away from our dependency pointing to this repo's master (or the specific commit rather) if I can.

And then one final issue running the tests - `test prop::test_get_policykit_version ... FAILED` which fails because there is no system dbus on my mac. Everything else passed though.

### DIFF
--- a/dbus/examples/adv_server.rs
+++ b/dbus/examples/adv_server.rs
@@ -70,7 +70,7 @@ fn create_iface(check_complete_s: mpsc::Sender<i32>) -> (Interface<MTFn<TData>, 
             })
             .on_set(|i, m| {
                 let dev: &Arc<Device> = m.path.get_data();
-                let b: bool = try!(i.read());
+                let b: bool = i.read()?;
                 if b && dev.checking.get() {
                     return Err(MethodErr::failed(&"Device currently under check, cannot bring online"))
                 }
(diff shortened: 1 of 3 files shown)


diwic:
Hi again and thanks for the PR! 

To start from the bottom, if not having a system dbus (or no policykit on it) is the common practice on mac, then I guess you could submit a separate PR that just skips the entire test on macs, preferrably with a comment explaining why.

diwic:
As for the rest, the second commit looks okay  to merge to master and then release a final 0.6 version.

The first commit is mostly 'try' to '?' changes, which I've already made in the 0.7 branch.

diwic:
Ok, I have now done this and released a v0.6.5. Enjoy!

Issue #195, opened 2019-07-30
What do you think about replacing timeout_ms on the std::time::Duration


a1ien:
In the code timeout is used as i32 value. But meybe we can cange api to use std::time::Duration?

diwic:
Sounds good, we should at least do this for the `blocking` module. For `ffidisp` let's leave it unchanged for backwards compatibility. Do you want to make a PR?

a1ien:
Ok. I make PR. But 0.7 already have breaking changes. Why not change also in ffidisp?

diwic:
I think of `ffidisp` primarily as a legacy module for people who don't want to migrate to `blocking` at this point. All they have to do then is to change imports.

a1ien:
Ohh. I forgot about arg::messageitem::Props also replace?

diwic:
Naah, I should probably deprecate `Props` altogether instead...

PR #196, opened 2019-07-31
Replacing timeout_ms on the std::time::Duration

a1ien:
Replace only blocking module. ffidisp is a legacy module for easy migration.

Closes: #195


diwic:
Looks good. Thanks!

PR #197, opened 2019-07-31
Fix Arg, Append, Get, RefArg implement for HashMap

a1ien:
Now we can use map with custom hasher.


diwic:
What difference does this make?

a1ien:
Without this HashMap or HashSet with custom hashers cannot be used with them.
For example we cant push or get fxhash as args

PR #198, opened 2019-07-31
Some suggest from clippy

a1ien:


a1ien:
I run `cargo check`. `cargo test` fails in `dbus-tokio`.

```
error[E0277]: `(dyn for<'r> std::ops::FnOnce(&'r mut dbus::Message) -> std::result::Result<_, dbus::Error> + 'static)` cannot be sent between threads safely
  --> dbus-tokio/src/connection.rs:69:19
   |
69 |     let fut = fut.boxed().compat();
   |                   ^^^^^ `(dyn for<'r> std::ops::FnOnce(&'r mut dbus::Message) -> std::result::Result<_, dbus::Error> + 'static)` cannot be sent between threads safely
   |
   = help: the trait `std::marker::Send` is not implemented for `(dyn for<'r> std::ops::FnOnce(&'r mut dbus::Message) -> std::result::Result<_, dbus::Error> + 'static)`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<(dyn for<'r> std::ops::FnOnce(&'r mut dbus::Message) -> std::result::Result<_, dbus::Error> + 'static)>`
   = note: required because it appears within the type `std::boxed::Box<(dyn for<'r> std::ops::FnOnce(&'r mut dbus::Message) -> std::result::Result<_, dbus::Error> + 'static)>`
   = note: required because it appears within the type `std::option::Option<std::boxed::Box<(dyn for<'r> std::ops::FnOnce(&'r mut dbus::Message) -> std::result::Result<_, dbus::Error> + 'static)>>`
   = note: required because it appears within the type `dbus::nonblock::MethodReply<_, std::sync::Arc<dbus::nonblock::Connection>>`

```
Without this test all test passed.

diwic:
Ok, I fixed up the dbus-tokio error, thanks for noticing.

I merged this one, but in general I'm hesitant to fix up things that are just clippy suggestions - mainly for the "don't fix what ain't broken" - and just because clippy suggests them doesn't mean that they are necessarily better, it's just a matter of clippy's developer's taste...

PR #200, opened 2019-08-01
Add travis-ci

a1ien:
For check that PR compile.

You need just enable https://travis-ci.com/
Status for this PR
https://travis-ci.com/a1ien/dbus-rs/builds/121546152
By now travis have bug. And with use ubuntu 16.04 instead 18.04. When this fix we can uncomment last line for run test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: rust
+dist: xenial
+
+cache: cargo
+
+before_install:
+  - sudo apt-get install -y libdbus-1-dev
+
+before_script:
+  - export DBUS_SESSION_BUS_ADDRESS=`dbus-daemon --session --print-address --fork`
+
+script:
+  - cargo update
+  - cargo check --all
+  - cargo test --all-features --all -- --nocapture --color always
--- a/dbus/src/arg/messageitem.rs
+++ b/dbus/src/arg/messageitem.rs
@@ -778,23 +778,23 @@ mod test {
         c.send(m).unwrap();
     }
 
-    /* Unfortunately org.freedesktop.DBus has no properties we can use for testing, but PolicyKit should be around on most distros. */
+    /* Unfortunately org.freedesktop.DBus has no properties we can use for testing, but hostname1 should be around on most distros. */
     #[test]
-    fn test_get_policykit_version() {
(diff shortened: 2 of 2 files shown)


a1ien:
I thought about turning it on cash. But since we still have one simple, I don’t see a big advantage. But I can add.

Ichimonji10:
I just did a quick empirical test to see how caching affects certain commands on one of my own personal projects:

* `cargo check`: 6s vs 1m 12s
* `cargo test`: 16s vs 1m 17s

See [with caching](https://travis-ci.org/Ichimonji10/killjoy/jobs/566252442#L235-L243) and [without caching](https://travis-ci.org/Ichimonji10/killjoy/jobs/566641803#L238-L367).

I didn't compare the total build time with and without caching, as in my experience, Travis CI build times tend to vary quite a bit.

diwic:
Is it possible to run `cargo test` now that the `dbus-tokio` error is fixed up (and why are default features disabled?)

a1ien:
We can't run `cargo test` now. Because now travis uses the wrong VM for rust.
https://travis-ci.community/t/for-dist-bionic-xenial-vm-is-used-instead-with-language-rust/4487
And we need 1.12 libdbus for `dbus_try_get_local_machine_id` but xenial use 1.10. Abd bionic have 1.12 dbus.

Log with enabled `cargo test` on travis

a1ien:
Removed `no-default-features`. They not need in our case.

diwic:
Ok. Actually I want the code to run on xenial so I'll look into if it's possible to not use `dbus_try_get_local_machine_id`.

diwic:
Ok, I just pushed a commit that makes us use `dbus_get_local_machine_id` instead. Want to try again?

a1ien:
Now all test passed on travis
https://travis-ci.com/a1ien/dbus-rs/builds/121817547
I change PolicyKit to hostname1 for Props test

diwic:
Okay...so I tried setting this up...and it worked fine at first, then I coded some more and now the CI is broken again, because the environment in the CI is different from my working machine. 

Any idea of a D-Bus service that implements Peer on the travis CI?

PR #201, opened 2019-08-01
Make BusType hashable

Ichimonji10:
Doing this makes it possible to deduplicate instances of `BusType` with
a `HashSet`. This change is also in line with the C-COMMON-TRAITS Rust
API guideline.

See:
https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits

### DIFF
--- a/dbus/src/channel.rs
+++ b/dbus/src/channel.rs
@@ -25,6 +25,7 @@ impl Drop for ConnHandle {
 }
 
 /// Which bus to connect to
+#[derive(PartialEq, Eq, Hash)]
 pub enum BusType {
     /// The Session bus - local to every logged in session
     Session = ffi::DBusBusType::Session as isize,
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I have an application which iterates through a series of objects and finds all
unique `BusType` instances. It could be implemented with something like this:

```rust
// Get a deduplicated list of D-Bus bus types in the given list of rules.
pub fn get_bus_types(rules: &[Rule]) -> Vec<BusType> {
    rules
        .iter()
        .map(|rule: &Rule| encode_bus_type(rule.bus_type))
        .collect::<HashSet<BusType>>()
}

```

The lack of hashability forces me to implement a workaround:

```rust
// Get a deduplicated list of D-Bus bus types in the given list of rules.
pub fn get_bus_types(rules: &[Rule]) -> Vec<BusType> {
    // The conversion from BusType → String → BusType is a hack. It's done because this method
    // should deduplicate BusType values, but BusType doesn't implement the traits necessary to
    // create a HashSet<BusType>.
    rules
        .iter()
        .map(|rule: &Rule| encode_bus_type(rule.bus_type))
        .collect::<HashSet<String>>()
        .into_iter()
        .map(|bus_type_str: String| decode_bus_type_str(&bus_type_str[..]).unwrap())
        .collect()
}

```

(Yes, my use of `String` instead of e.g. `usize` is inefficient. I'll probably refactor.)

PR #202, opened 2019-08-02
Use TryFrom instead custom FromMessageItem

a1ien:


diwic:
I guess this is okay, but why are you spending time on something as old and obsolete as MessageItem?

a1ien:
I did not know that it is obsolete :)

diwic:
Tbh, I don't see much advantage in doing this, but OTOH, I don't see much disadvantage either, so if it makes you happy...merged now 🙂

PR #204, opened 2019-08-03
Add dyn keyword.

a1ien:
`Box<Trait>` will become deprecated in favor of using `Box<dyn Trait>`

If we try to build on nightly we see many warnings


diwic:
Since you like refactoring things...

I think every place that looks like this: `(&var as &Trait).method(...)` should instead be changed to `Trait::method(&var, ...)`. Because we don't want dynamic dispatch where we can avoid it and the addition of `dyn` makes this less clear.

The rest seems okay to me.

PR #217, opened 2019-10-08
dbus-codegen: Add filter option + Add command line to generated header

davidgraeff:
Only generate interfaces that match given filters.

Usage: --filter Interface1,Interface2
Example: dbus-codegen-rust -d org.freedesktop.DBus -p /org/freedesktop/DBus -m none -c nonblock -g -i org.freedesktop. --filter DBus

The generated header contains the command arguments as well:

```
// This code was autogenerated with `dbus-codegen-rust -d org.freedesktop.DBus -p /org/freedesktop/DBus -m none -c nonblock -g -i org.freedesktop. --filter DBus`, see https://github.com/diwic/dbus-rs
```

This PR allows me to auto-generate (without further fine tuning) the `nonblock::stdintf::dbus` interface for a follow up PR to add add_match to the non blocking / async api.

Signed-off-by: David Gräff <david.graeff@web.de>


diwic:
Hi and thanks for the PR!

If you want to auto-generate the `stdintf.rs` files, you have an additional thing to solve: I hand-edit the `Properties` interface to descend into the variants; I never got around to solving generating this automatically.

dbus-codegen/src/generate.rs
```diff
@@ -80,13 +81,16 @@ pub struct GenOpts {
     pub futures: bool,
     /// Type of connection, for client only
     pub connectiontype: ConnectionType,
+    /// trait filter. Only matching traits are generated, if non-empty.
```
diwic: Call this "interface filter" instead of "trait filter", as it matches on the D-Bus interface name, not the generated trait name (as it should be).

It should be an `Option<HashSet<String>>` so that `None` matches everything and `Some(HashSet::new())` matches nothing.

Also, a better name for the variable would be `interfaces` or  `interface_filter`.

dbus-codegen/src/generate.rs
```diff
@@ -663,12 +668,12 @@ fn write_module_header(s: &mut String, opts: &GenOpts) {
 }
 
 /// Generates Rust structs and traits from D-Bus XML introspection data.
-pub fn generate(xmldata: &str, opts: &GenOpts) -> Result<String, Box<dyn error::Error>> {
+pub fn generate(xmldata: &str, opts: &GenOpts, command_line: String) -> Result<String, Box<dyn error::Error>> {
     use xml::EventReader;
```
diwic: `command_line` should be inside `GenOpts` instead of an additional parameter.

dbus-codegen/src/generate.rs
```diff
@@ -691,6 +696,11 @@ pub fn generate(xmldata: &str, opts: &GenOpts) -> Result<String, Box<dyn error::
                 if curm.is_some() { Err("End of Interface inside method")? };
                 if curintf.is_none() { Err("End of Interface outside interface")? };
                 let intf = curintf.take().unwrap();
+                // If filters are set and no filter matches -> Just print a message and continue parsing
+                if !opts.filter.is_empty() && !opts.filter.contains(&intf.shortname) {
+                    eprintln!("Skip filtered interface '{}'", &intf.shortname);
+                    continue;
```
diwic: The `eprintln` there might be useful for debugging during development, but for the end user it's probably more of a distraction, so delete it or comment it out.

dbus-codegen/src/main.rs
```diff
@@ -34,6 +34,9 @@ fn main() {
              .help("If present, connects to the supplied service to get introspection data. Reads from stdin otherwise."))  
         .arg(clap::Arg::with_name("path").short("p").long("path").takes_value(true).value_name("PATH")
              .help("The path to ask for introspection data. Defaults to '/'. (Ignored if destination is not specified.)"))
+        .arg(clap::Arg::with_name("filter").short("f").long("filter").takes_value(true).value_name("FILTER")
+            .help("Comma separated list of filter strings. Only matching traits are generated. \
```
diwic: should probably be named `--interfaces` instead of `--filter`, this would be more clear, especially in case I want to add more filters in the future.

diwic:
An additional thought is whether the list of interfaces should just be a comma separated list, or if it should actually be a regex?

davidgraeff:
Thanks for the review time.

> I hand-edit the Properties interface

Yes I've noticed this. You are adding `'static`s and you remove the boxing `Variant` from the `get` and `get_all` method. I do not see how that can be identified automatically without prior knowledge of the dbus response type. I thought about adding a "special_treatment" map type to encode such changes ... or just hack them in at appropriate places for just the dbus properties type.

> ...and you can't just delete nonblock/stdintf.rs

Oh I did in my personal test branch. The user interface stays the same, I think. You have some module nesting going in `nonblock.rs` on to mimic the current API, but you end up with a fully generated directory. Sounds clean to me, I'm curious to know your objections. The code in question looks like this:

```rust
mod generated_org_freedesktop_notifications;
mod generated_org_freedesktop_dbus;

pub mod stdintf {
    pub mod org_freedesktop_dbus {
        pub use super::super::generated_org_freedesktop_notifications::*;
        pub use super::super::generated_org_freedesktop_dbus::*;
    }
}
```

I agree with the rest of your review, although regarding the `eprintln` I liked some feedback on which classes are filtered out.

> or if it should actually be a regex?

I thought about it as well. On the other hand, we are dealing with about 3-5 interfaces on average per dbus tree node. Interfaces appear in a repetitive pattern on a dbus node hierarchy and generating an interface is done by hand usually only a few times ever. Is it worth to add another dependency here?

diwic:
> You are adding 'statics and you remove the boxing Variant from the get and get_all method. I do not see how that can be identified automatically without prior knowledge of the dbus response type.

I think it could be identified automatically. Maybe I'll try to fix it as I'm getting tired of the hand editing as well...

I'm okay with keeping the `eprintln` for the time being (i e, until someone complains...) and keeping it as a comma separated list.

>>    ...and you can't just delete nonblock/stdintf.rs
> Oh I did in my personal test branch.

The PR you submitted does indeed delete `nonblock/stdintf.rs` without providing a replacement - the code that deletes this file and the code that provides a replacement needs to be in the same PR 🙂

diwic:
With https://github.com/diwic/dbus-rs/commit/bfc4737b3647c34f5f66deabcd26579ae409347c you should have one obstacle less to generate stdintf automatically. Give it a try!

davidgraeff:
> The PR you submitted does indeed delete nonblock/stdintf.rs without providing a replacement 

D'oh! Not intentional. The generated files work perfectly now. I'll add them to this PR in a few moments.

diwic:
Okay, I think this is ready for merging if you
 * change the `generated_org_freedesktop_dbus` module to not be visible outside the crate, as the plan is to write wrappers on the connection like we do for the `blocking` module. At least for now.
 * squash it into two commits (one for the changes in the dbus-codegen dir, one for the changes in the dbus dir)

davidgraeff:
I have squashed the commits. I tend to not agree with you on hiding the Dbus trait completely and have not made any changes yet. The exercise of that change is to allow me to use Dbus::add_match in nonblocking mode. 

What about making raw interfaces available in a sub module "raw" or "unstable"?

diwic:
I'm not sure I'm following - you were planning to add an `add_match` function on `nonblock::LocalConnection`, right? In that case crate visibility shouldn't be a problem.

davidgraeff:
> you were planning to add an add_match function

Absolutely. It is in no way convenient to use the raw `add_match` method of the DBus interface.
You do have the generated `DBus` trait public in the blocking module:
https://github.com/diwic/dbus-rs/blob/master/dbus/src/blocking/stdintf.rs#L165

Just wondering and trying to understand. If the plan is to make that an implementation detail in a future version iteration then it totally makes sense to hide it right away for nonblocking.

diwic:
Okay, so the reason the `DBus` is inside a `pub (super)`, not `pub` module in `blocking` (look again!) is this:

    pub struct DBusNameOwnerChanged {
        pub arg0: String,
        pub arg1: String,
        pub arg2: String,
    }

I don't want `arg0` in the public API.

dbus/src/nonblock.rs
```diff
@@ -16,7 +16,20 @@ use std::{future, task, pin, mem};
...
+
+/// This module contains some standard interfaces and an easy way to call them.
+///
+/// See the [D-Bus specification](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces) for more information about these standard interfaces.
+///
+/// The code was created by dbus-codegen.
+pub mod stdintf {
+    pub mod org_freedesktop_dbus {
+        pub use super::super::generated_org_freedesktop_notifications::*;
+        pub(crate) use super::super::generated_org_freedesktop_dbus::*;
```
davidgraeff: It's `pub(crate)` now

dbus/src/nonblock/generated_org_freedesktop_dbus.rs
```diff
@@ -0,0 +1,209 @@
+// This code was autogenerated with `dbus-codegen-rust -d org.freedesktop.DBus -p /org/freedesktop/DBus -m none -c nonblock -g -i org.freedesktop. --dbuscrate crate --filter DBus -o dbus/src/nonblock/generated_org_freedesktop_dbus.rs`, see https://github.com/diwic/dbus-rs
+#![allow(missing_docs)]
```
davidgraeff: This is currently not automatically added. It fails for me for files that are `include!(..)`d, like in the dbus-codegen tests. But it is generally useful for those auto generated files.

Should this be a generator option? (defaulting to add `#![allow(missing_docs)]`)

diwic: Another option could be to actually insert some docs, like "This is the `Foo` signal on the `bar.baz` interface".

davidgraeff: Those would be very generic. A dbus introspection response may actually have annotations for this purpose, you probably know that already. But no service is really using them. The question is, if it is worth parsing those annotations and what to do with all the non-annotated methods and interfaces.

PR #219, opened 2019-10-13
Update Argument Guide examples

parrotmac:
Hey there!

Thanks for the project, this looks really useful :slightly_smiling_face:. It looks like the API changed at some point, so this just updates one of the examples so it can be compiled.


diwic:
Thanks!

PR #220, opened 2019-10-17
Add MessageItem::peel() and use it in TryFrom impl's

danieldulaney:
This PR adds `MessageItem::peel()`, a method that returns a reference to the outermost `MessageItem` that isn't a `Variant`. This utility function is useful when you just care about the content of the item without paying attention to exactly how it's structured. The first commit (18fd78e) adds this method.

This PR also adds peel() to the `TryFrom` `impl`'s (and therefore adds it to `MessageItem::inner()`). This reflects the user's intention when using `inner()`: that this is the true content of the `MessageItem`, regardless of structure. However, this is a minor breaking change: `MessageItem::Variant(...).inner()` that produced an `Err(())` previously now returns `Ok(_)`, which is slightly different behavior. This behavior comes in the second commit (0c3ac1b), which can be submitted separately if necessary.

Tests are included.

### DIFF
--- a/dbus/src/arg/messageitem.rs
+++ b/dbus/src/arg/messageitem.rs
@@ -332,7 +332,7 @@ macro_rules! msgitem_convert {
         impl<'a> TryFrom<&'a MessageItem> for $t {
             type Error = ();
             fn try_from(i: &'a MessageItem) -> Result<$t,()> {
-                if let MessageItem::$s(b) = i { Ok(*b) } else { Err(()) }
+                if let MessageItem::$s(b) = i.peel() { Ok(*b) } else { Err(()) }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


diwic:
I've added a commit now that make `OwnedFd::new` unsafe.

danieldulaney:
Good call on the `OwnedFd` -- I'll make that change.

As for why, I'm actually working on a PR that brings notify-rust up to dbus 0.7, and it pretty heavily uses `MessageItem`. With the change from `DictItem`s to `Dict`s, some of the `MessageItem`-based code needed to be reworked, but moving everything to the new API is a longer-term effort.

diwic:
> However, without changing the OwnedFd API, there's nothing you can actually do with an &OwnedFd.

You can do `as_raw_fd`...

Anyhow, it looks good to me now, so I'm merging it. Although I'd still encourage you to move things over to `Arg` (and `RefArg`) 🙂

PR #224, opened 2019-12-16
Updated tokio to release 0.2

paulusx:
Updated tokio to release. Fixed some clippy warnings

### DIFF
--- a/dbus-tokio/src/connection.rs
+++ b/dbus-tokio/src/connection.rs
@@ -32,7 +32,9 @@ impl<C: AsRef<Channel> + Process> IOResource<C> {
                 self.registration = Some((reg, w.fd));
             }
             Some((_reg, fd)) if *fd == w.fd => (),
-            Some((_reg, _fd)) => unreachable!(),
+            Some((_reg, fd)) => return Err(Box::new(Error::new_failed(
+                &format!("Unexpected changing file descriptor for dbus (from {} to {})",  w.fd, fd)
+            ))),
         };
 
         if let Some((r,_)) = &self.registration {
--- a/dbus-tokio/src/connection.rs
+++ b/dbus-tokio/src/connection.rs
@@ -83,19 +83,26 @@ pub fn new_system() -> Result<(IOResource<Connection>, Arc<Connection>), Error> 
 
 #[test]
 fn method_call() {
+    use tokio::task;
+    use futures::future::FutureExt;
+
     let mut rt = tokio::runtime::Builder::new()
         .basic_scheduler()
         .enable_io()
         .build()
         .unwrap();
 
-    let (res, conn) = new_session_sync().unwrap();
-    rt.spawn(async { panic!(res.await);});
+    let local = task::LocalSet::new();
+
+    let (res, conn) = new_session_local().unwrap();
+    let spawner = async move {
+        task::spawn_local(async move { panic!(res.await);});
+    };
 
     let proxy = dbus::nonblock::Proxy::new("org.freedesktop.DBus", "/", conn);
     let fut = proxy.method_call("org.freedesktop.DBus", "NameHasOwner", ("dummy.name.without.owner",));
 
-    let (has_owner,): (bool,) = rt.block_on(fut).unwrap();
+    let (has_owner,): (bool,) = local.block_on(& mut rt , spawner.then(|_| fut)).unwrap();
 
     assert_eq!(has_owner, false);
 }
--- a/dbus-tokio/src/connection.rs
+++ b/dbus-tokio/src/connection.rs
@@ -5,7 +5,7 @@ use dbus::Error;
 use std::{future, task, pin};
 use std::sync::Arc;
 
-use tokio_net::driver::Registration;
+use tokio::io::Registration;
 
 /// The I/O Resource should be spawned onto a Tokio compatible reactor.
 ///
@@ -14,31 +14,44 @@ use tokio_net::driver::Registration;
 /// contact with the D-Bus server.
 pub struct IOResource<C> {
     connection: Arc<C>,
-    registration: Registration,
+    registration: Option<(Registration, std::os::unix::io::RawFd)>,
 }
 
 impl<C: AsRef<Channel> + Process> IOResource<C> {
-    fn poll_internal(&self, ctx: &mut task::Context<'_>) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    fn poll_internal(&mut self, ctx: &mut task::Context<'_>) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         let c: &Channel = (*self.connection).as_ref();
 
         c.read_write(Some(Default::default())).map_err(|_| Error::new_failed("Read/write failed"))?;
         self.connection.process_all();
 
         let w = c.watch();
-        let r = &self.registration;
-        r.register(&mio::unix::EventedFd(&w.fd))?;
-        r.take_read_ready()?;
-        r.take_write_ready()?;
 
-        if w.read { let _ = r.poll_read_ready(ctx)?; };
-        if w.write { let _ = r.poll_write_ready(ctx)?; };
+        match &mut self.registration {
+            None => {
+                let reg = Registration::new(&mio::unix::EventedFd(&w.fd))?;
+                self.registration = Some((reg, w.fd));
+            }
+            Some((_reg, fd)) if *fd == w.fd => (),
+            Some((_reg, _fd)) => unreachable!(),
+        };
+
+        if let Some((r,_)) = &self.registration {
+            r.take_read_ready()?;
+            r.take_write_ready()?;
+
+            if w.read { let _ = r.poll_read_ready(ctx)?; };
+            if w.write { let _ = r.poll_write_ready(ctx)?; };
+        } else {
+            unreachable!()
+        }
 
         Ok(())
+
     }
 }
 
 impl<C: AsRef<Channel> + Process> future::Future for IOResource<C> {
-    fn poll(self: pin::Pin<&mut Self>, ctx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
+    fn poll(mut self: pin::Pin<&mut Self>, ctx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
         match self.poll_internal(ctx) {
             Ok(_) => {
                 task::Poll::Pending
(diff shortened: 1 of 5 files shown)


diwic:
Hi and thanks for the PR! Looks good mostly (to the best of my knowledge), just one thing:

You have changed the test code from `new_session_local` to `new_session_sync`, I'd like to test the local session together with a thread-local runtime like it was previously. Is that possible?

paulusx:
Hi! Thank you for your crate!:) I have tried to make it, but the following error have occurred: ```
error[E0277]: `std::cell::RefCell<dbus::filters::Filters<std::boxed::Box<(dyn for<'r> std::ops::FnMut(dbus::message::Message, &'r dbus::nonblock::LocalConnection) -> bool + 'static)>>>` cannot be shared between threads safely
  --> dbus-tokio/src/connection.rs:93:8
   |
93 |     rt.spawn(async { panic!(res.await);});
   |        ^^^^^ `std::cell::RefCell<dbus::filters::Filters<std::boxed::Box<(dyn for<'r> std::ops::FnMut(dbus::message::Message, &'r dbus::nonblock::LocalConnection) -> bool + 'static)>>>` cannot be shared between threads safely
   |
```

**Comment by paulusx** [comment]
> Hi and thanks for the PR! Looks good mostly (to the best of my knowledge), just one thing:
> 
> You have changed the test code from `new_session_local` to `new_session_sync`, I'd like to test the local session together with a thread-local runtime like it was previously. Is that possible?

I fixed it

File: dbus-tokio/src/connection.rs [hunk 1]
```diff
@@ -14,31 +14,44 @@ use tokio_net::driver::Registration;
...
+        match &mut self.registration {
+            None => {
+                let reg = Registration::new(&mio::unix::EventedFd(&w.fd))?;
+                self.registration = Some((reg, w.fd));
+            }
+            Some((_reg, fd)) if *fd == w.fd => (),
+            Some((_reg, _fd)) => unreachable!(),
+        };
+
+        if let Some((r,_)) = &self.registration {
```
davidgraeff: I suggest to use `expect` (or unwrap) here. Right now the compiler does not get the chance to optimize the branch away.

In my copy of this file at https://github.com/openhab-nodes/wifi-captive-rs/blob/master/src/network_interface/connection.rs#L31 I have set the `registry` field during initialization (see https://github.com/openhab-nodes/wifi-captive-rs/blob/master/src/network_interface/connection.rs#L65), because in the current dbus implementation the file descriptor never changes.

dbus-tokio/src/connection.rs
```diff
@@ -51,12 +64,13 @@ impl<C: AsRef<Channel> + Process> future::Future for IOResource<C> {
 
 
 /// Generic connection creator, you might want to use e g `new_session_local`, `new_system_sync` etc for convenience.
-pub fn new<C: From<Channel>>(b: BusType) -> Result<(IOResource<C>, Arc<C>), Error> {
+pub fn new< C: From<Channel>>(b: BusType) -> Result<(IOResource< C>, Arc<C>), Error> {
     let mut channel = Channel::get_private(b)?;
```
diwic: Weird whitespace change?

dbus-tokio/src/connection.rs
```diff
@@ -69,16 +83,26 @@ pub fn new_system() -> Result<(IOResource<Connection>, Arc<Connection>), Error>
...
+        .build()
+        .unwrap();
+
+    let local = task::LocalSet::new();
 
     let (res, conn) = new_session_local().unwrap();
-    rt.spawn(async { panic!(res.await); });
+    let spawner = async move {
+        task::spawn_local(async move { panic!(res.await);});
+    };
```
diwic: This looks a bit superfluous? Look at the example here: https://docs.rs/tokio/0.2.4/tokio/task/struct.LocalSet.html#method.spawn_local

Any reason we can't simplify in the same way?

dbus/src/message/signalargs.rs
```diff
@@ -44,6 +44,7 @@ pub trait SignalArgs {
     /// If the message is a signal of the correct type, return its arguments, otherwise return None.
     ///
     /// This does not check sender and path of the message, which is likely relevant to you as well.
+    #[allow(clippy::if_same_then_else)]
     fn from_message(m: &Message) -> Option<Self> where Self: Sized + arg::ReadAll {
```
diwic: Maybe we should instead rewrite the function like:

    if a { return None };
    if b { return None };

...instead of adding clippy lints in the code.

dbus-tokio/src/connection.rs
```diff
@@ -69,16 +83,26 @@ pub fn new_system() -> Result<(IOResource<Connection>, Arc<Connection>), Error>
...
-    use tokio::runtime::current_thread::Runtime;
+    use tokio::task;
+    use futures::future::FutureExt;
 
-    let mut rt = Runtime::new().unwrap();
+    let mut rt = tokio::runtime::Builder::new()
+        .basic_scheduler()
+        .enable_io()
+        .build()
+        .unwrap();
```
diwic: Why is this change needed?

dbus-tokio/src/connection.rs
diwic: Agreed.

paulusx: @davidgraeff Thanks! Your solution looks better:) I have tried copy-paste it, but see this runtime error in test: 

```running 1 test
test connection::method_call ... FAILED

failures:

---- connection::method_call stdout ----
thread 'connection::method_call' panicked at '`spawn_local` called from outside of a task::LocalSet!', src/libcore/option.rs:1190:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    connection::method_call
```

davidgraeff:
@diwic Could you remove the other files of this crate after this PR, maybe even remove the dbus-futures crate and release a first tokio-0.2 based alpha? I know that you probably wanna have crossroads included before a final release. But the connection class alone is already useful, as seen by my other project.

You may also want to have a look at https://github.com/openhab-nodes/wifi-captive-rs/blob/master/src/network_interface/signal_stream.rs, my idea for an async dbus signal watcher based on a tokio stream (not entirely generic in my implementation though). This would complement the dbus-tokio crate.

paulusx:
https://github.com/diwic/dbus-rs/pull/224#discussion_r358324352
If call ` task::spawn_local` outside runtime (not in future) — I see error: `thread 'connection::method_call' panicked at 'no current reactor`.  I can't found `lazy` in current *tokio*

diwic:
> [#224 (comment)](https://github.com/diwic/dbus-rs/pull/224#discussion_r358324352)
> If call ` task::spawn_local` outside runtime (not in future) — I see error: `thread 'connection::method_call' panicked at 'no current reactor`. I can't found `lazy` in current _tokio_

"Unlike the free function spawn_local, TaskSet::spawn_local may be used to spawn local tasks when the task set is not running."

paulusx:
Sorry. I'm newbie in tokio. I haven't found `TaskSet` in tokio. Maybe you mean 'tokio::task::LocalSet`? I read docs, but when I have written function as:

```
#[test]
fn method_call() {
    use tokio::task;
    use futures::future::FutureExt;

    let mut rt = tokio::runtime::Builder::new()
        .basic_scheduler()
        .enable_io()
        .build()
        .unwrap();

    let local = task::LocalSet::new();

    let (res, conn) = new_session_local().unwrap();
    local.spawn_local(async move { panic!(res.await);});

    let proxy = dbus::nonblock::Proxy::new("org.freedesktop.DBus", "/", conn);
    let runner = proxy.method_call("org.freedesktop.DBus", "NameHasOwner", ("dummy.name.without.owner",));

    let (has_owner,): (bool,) = local.block_on(& mut rt , runner).unwrap();

    assert_eq!(has_owner, false);
}
``` 
then I saw the error 

```
---- connection::method_call stdout ----
thread 'connection::method_call' panicked at 'no current reactor', /home/paulus/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.4/src/io/driver/mod.rs:243:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

diwic:
> @diwic Could you remove the other files of this crate after this PR, maybe even remove the dbus-futures crate and release a first tokio-0.2 based alpha? I know that you probably wanna have crossroads included before a final release. But the connection class alone is already useful, as seen by my other project.

@davidgraeff 
I've been a bit busier than expected lately, which is why crossroads hasn't made much progress last few weeks. Now that I'm thinking about it, maybe the best course of action is to make it into its own crate, at least for the time being. That way I can iterate on dbus-rs and dbus-crossroads separately.


> You may also want to have a look at https://github.com/openhab-nodes/wifi-captive-rs/blob/master/src/network_interface/signal_stream.rs, my idea for an async dbus signal watcher based on a tokio stream (not entirely generic in my implementation though). This would complement the dbus-tokio crate.

Okay, is the code licensed under Apache 2.0/MIT so I can copy-paste from it or should I study and rewrite?

diwic:
Okay, I looked it up. It's `Registration::new` that must be called from inside the reactor. We then have two choices:

 1) Change the code so that `Registration::new` is called from within poll / poll_internal or
 2) Require that `new_session_local` and friends is called from inside the reactor.

I strongly prefer the first option. The the test code can be simplified to:

    let local = task::LocalSet::new();

    let (res, conn) = new_session_local().unwrap();
    local.spawn_local(async move { panic!(res.await);});

    let proxy = dbus::nonblock::Proxy::new("org.freedesktop.DBus", "/", conn);
    let (has_owner,): (bool,) = local.block_on(&mut rt, 
        proxy.method_call("org.freedesktop.DBus", "NameHasOwner", ("dummy.name.without.owner",))
    ).unwrap();

    assert_eq!(has_owner, false);

davidgraeff:
Guys, you know about https://docs.rs/tokio/0.2.4/tokio/attr.test.html right?
Those macros (#[tokio::test(threaded_scheduler)] and #[tokio::test]) set up the reactor without much fiddling.

diwic:
@davidgraeff  In this case I think it's good that we are not using it, since that helped us discover a problem with the solution of having `Registration::new` inside the session constructor.

davidgraeff:
True. I have checked the documentation of "Registration::new" and indeed this constructor method does not only create a Registration type but also associates it with the default reactor.

How about making the method async? That way the API expresses the need for a runtime.
This would be more like option 2 though.

diwic:
No, that has other limitations. E g, some people may want to call `new_session_sync` on a background thread since it is blocking, and then spawn the IOResource on a different reactor.

paulusx:
I have reverted *Registration* in *poll_internal*

diwic:
Ok, I decided that it was easiest to merge the patch and then fixup the remaining things myself (https://github.com/diwic/dbus-rs/commit/9325c6aa0977fccd1459f88d49bb7b20b8cce14a). Thanks for helping out!

davidgraeff:
Hm. Seems inefficient with those many branches in the hotpath (one for unpacking the option and a second one to compare the file descriptor) if actual file descriptor changes of dbus are not even allowed in this implementation.

diwic:
@davidgraeff Feel free to make a separate PR if you want to improve something.

paulusx:
Thank for reviews)

PR #237, opened 2020-02-23
MatchRule: Add "eavesdrop" bool flag

xkr47:
This allows implementing e.g. dbus-monitor in Rust.
See https://stackoverflow.com/questions/11544836/monitoring-dbus-messages-by-python

I could add a simple dbus-monitor implementation to the examples if you like.

### DIFF
--- a/dbus/src/message/matchrule.rs
+++ b/dbus/src/message/matchrule.rs
@@ -21,6 +21,8 @@ pub struct MatchRule<'a> {
     pub interface: Option<Interface<'a>>,
     /// Match on message member (signal or method name)
     pub member: Option<Member<'a>>,
+    /// If true, also receive messages not intended for us. Defaults to false.
+    pub eavesdrop: bool,
     _more_fields_may_come: (),
 }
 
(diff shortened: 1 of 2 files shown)


dbus/src/message/matchrule.rs
```diff
@@ -45,6 +47,7 @@ impl<'a> MatchRule<'a> {
         if let Some(ref x) = self.path { v.push((pn, &x)) };
         if let Some(ref x) = self.interface { v.push(("interface", &x)) };
         if let Some(ref x) = self.member { v.push(("member", &x)) };
+        if self.eavesdrop { v.push((pn, "true")) };
```
diwic: you probably mean "eavesdrop" instead of `pn`

diwic:
> I could add a simple dbus-monitor implementation to the examples if you like.

More examples are always welcome 🙂

xkr47:
Fixed. I threw in my example as well now.

diwic:
Looks good now. Thanks for contributing!

PR #238, opened 2020-02-23
README: Add subheading to make the "more examples" more prominent

xkr47:


xkr47:
Hmm you might have actually meant it to belong to the server directory? But up to you if you want to merge it anyway :)

diwic:
I decided to take your other PR, and reformat the server example section instead.

Not sure if `monitor.rs` is server or client but I think more client than server...

PR #240, opened 2020-02-23
Make some struct members public so we can actually read them

xkr47:
I don't know if you had better plans how to do it.

Actually I don't need the signatures, but I thought they too were stored there for them to be accessed :)

There were some more pub structs with non-pub members in the same file but didn't have time to investigate more.


xkr47:
Sorry I was in a hurry. I see the accessors now.. :/

PR #249, opened 2020-04-03
dbus: guarantee representations of arrays of 'simple' types.

### LINKED ISSUE #248: as_iter_of_u8 or similar method for RefArg
The [notification specification](https://developer.gnome.org/notification-spec/) indicates that images can be passed in the hints field of a message using a value of signature `(iiibiiay)`. Since the hint dict is untyped, at runtime when I'm implementing the server I only get a `HashMap<&str, Variant<Box<dyn RefArg>>>`. So I ultimately need to turn a `dyn RefArg` into a `Vec<u8>` or a `&[u8]` or similar... but as far as I can tell, the only way to do that is to use `as_iter`, which will box each individual byte of raw image data! I'd like to be able to just get an `Box<dyn Iterator<Item = u8>>` out. In theory this could be generalized to *any* of the types that can be stored in an array (byte, string, etc) but a `u8` is good enough for me for now.

(I saw that you're fairly uninterested in maintaining this these days; I don't think I could do full co-maintainership, but I'd definitely be interested in implementing this.)

deifactor:
Sometimes a large array might be stored behind a variant; for example, the 'show a notification' message in the desktop notifications spec can store an image in a dict of variants. So we should guarantee that you can cast that variant to a Vec of the right type.

We don't make that guarantee for more complicated types like structs or dicts because their representations are more up-in-the-air.

Fixes #248. Up to you if you like using the macros to automate the per-type checking in the tests, of course.

### DIFF
--- a/dbus/src/arg/msgarg.rs
+++ b/dbus/src/arg/msgarg.rs
@@ -64,12 +64,20 @@ pub trait RefArg: fmt::Debug {
     fn append(&self, _: &mut IterAppend);
     /// Transforms this argument to Any (which can be downcasted to read the current value).
     ///
-    /// Note: The internal representation of complex types (Array, Dict, Struct) is unstable
+    /// If the type is an array, and the element type is byte, numeric, boolean, or string, then the
+    /// representation is guaranteed to be a Vec<T>. Arrays of structs, dicts, and so on are not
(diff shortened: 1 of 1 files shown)


diwic:
Thanks!

PR #252, opened 2020-04-17
codegen: Remove the warnings on the generated code 

MusiKid:
Hi!
These modifications should remove some warnings on the code generated by `dbus-codegen`, with the `dyn` keyword and the unused `dbus::arg` module when the server code is generated with MethodInfo.


dbus-codegen/src/generate.rs
```diff
@@ -513,7 +513,7 @@ fn write_server_access(s: &mut String, i: &Intf, saccess: ServerAccess, minfo_is
             *s += "        let d = dd.as_ref();\n";
         },
         ServerAccess::RefClosure => *s += &format!("        let d = fclone({}minfo);\n", z),
-        ServerAccess::MethodInfo => *s += &format!("        let d: &{} = {}minfo;\n", make_camel(&i.shortname), z),
+        ServerAccess::MethodInfo => *s += &format!("        let d: &dyn {} = {}minfo;\n", make_camel(&i.shortname), z),
     }
```
diwic, line 517: Hmm, I guess the dynamic dispatch can be avoided but it's complicated so I think this is okay for now.

dbus-codegen/src/generate.rs
```diff
@@ -683,7 +683,9 @@ fn write_intf_crossroads(s: &mut String, i: &Intf, opts: &GenOpts) -> Result<(),
 fn write_module_header(s: &mut String, opts: &GenOpts) {
     *s += &format!("// This code was autogenerated with `dbus-codegen-rust {}`, see https://github.com/diwic/dbus-rs\n", opts.command_line);
     *s += &format!("use {} as dbus;\n", opts.dbuscrate);
-    *s += &format!("use {}::arg;\n", opts.dbuscrate);
+    if opts.serveraccess != ServerAccess::MethodInfo {
+        *s += &format!("use {}::arg;\n", opts.dbuscrate);
+    }
     if opts.futures {
```
diwic, line 689: This one is incorrect. The `arg` module is used in many places, such as `arg::Variant` or `arg::Get`.

dbus-codegen/src/generate.rs
MusiKid: I just wanted to fix the lint about [bare trait objects](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#bare-trait-object). Didn't know that it has implications other than to explicit the use of traits.

dbus-codegen/src/generate.rs
MusiKid: I wasn't sure about this one myself! I'm gonna remove it.

diwic:
Hi! 

So I applied your first commit and silenced the warning instead of applying your second commit. Could you check git master and see if all warnings are gone now for your use case?

MusiKid:
Yes it's working thanks!

PR #253, opened 2020-04-17
add option to run the dbus connection processing against a borrowed tree

icewind1991:
this removes the need to move the tree into a filter closure.
By allowing the user to keep ownership of the tree, it allows mutating the tree without having to use `Arc<Mutex<...>>` style tricks.

For an api usage pov, this replaces

```rust
tree.start_receive(&c);
loop { c.process(Duration::from_millis(1000))?; }
```

with

```rust
loop { c.process_with_tree(&tree, Duration::from_millis(1000))?; }
```

### DIFF
--- a/dbus/src/blocking.rs
+++ b/dbus/src/blocking.rs
@@ -127,6 +127,11 @@ impl $c {
             Ok(false)
         }
     }
+
+    /// The channel for this connection
+    pub fn channel(&self) -> &Channel {
+        &self.channel
+    }
 }
 
 impl BlockingSender for $c {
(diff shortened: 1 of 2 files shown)


diwic:
Hi and thanks for the PR!

I understand your argument but I think that for this use case you should rather use `Channel` directly.
Also the dependency should be the other way around, I don't want `Channel` to depend on `Tree`.

I e, could you come up with a method on `Tree` that takes a `&Channel` that does the same thing?

icewind1991:
done

diwic:
Looks good now. Thanks!

Issue #254, opened 2020-04-20
Failing to process large message replies with dbus-tokio


ivoisbelongtous:
I've been trying to use these bindings to interface with `connman`, but have been having issues when `connman` sends large payloads in reply to methods I've called, e.g., a long array of properties when querying for services. I suspect this is related to issues discussed in the initial tokio integration proposal, particularly https://github.com/diwic/dbus-rs/issues/72#issuecomment-312346002. The `dbus-tokio` implementation appears to have changed quite a lot since then (now using `dbus_connection_read_write` in favour of `dbus_connection_read_write_dispatch`) and so has `tokio` (the previous solution seems to have been switching to level triggers which is no longer an option with modern `tokio`.)

To summarise the linked issue and my own findings (corrections very welcome): the future gets woken up when D-Bus's watch object – represented by a file descriptor – is marked as ready for writing to or reading from. In this case, we care about whether we can read from D-Bus as this means there are messages to handle from the incoming queue. On waking up, the future calls `dbus_connection_read_write` to process the queue, then `dbus_connection_pop_message` to get the first message. However, when we receive a large message we run into a quirk where the watch is marked as ready but even after processing the queue still no message is popped. This seems to be because `dbus_connection_read_write` stops after 2048 bytes are processed, meaning the message is not actually ready to be dispatched, even if the watch claims it's ready to be read. The solution to this problem in the past was to set `tokio` to be level-triggered, meaning it would wake the future up whenever anything in the watch changed, spuriously waking up whenever any number of bytes were processed but ensuring we're awake to dispatch the final message. Unfortunately, `tokio` now only supports 'edge-triggering' which means the future will only be woken up when the watch first transitions to be read ready, and going back to sleep before the message is actually ready to dispatch (nor driving it to completion enough by calling `dbus_connection_read_write`.) This results in the reply never being handled and `Proxy::method_call` erroring with a timeout. (Big thanks to albel727 in the original issue who I won't @ directly so as not to rope them in)

I've created a minimal example to demonstrate the problem: https://github.com/ivoisbelongtous/dbus-large-message-timeout-example. Fiddling around with the number of entries and property fields in the method reply factory function should demonstrate that receiving the replies is dependent on the size of the payload. I've also confirmed this behaviour is only exhibited when using the `nonblock` module and not the `blocking` variation.

[...]

PR #255, opened 2020-04-21
test: Check for proper processing of large messages

ivoisbelongtous:
Tests for the issue raised in #254.

Not quite sure about the atomic ordering, but certainly looks similar to the example `Acquire`/`Release` use case.


diwic:
Thanks!

PR #258, opened 2020-04-30
Fixed several dbus-tokio freezes

Snarpix:
Hi! 
I had ran into spurious freezes that happens sometimes during startup in my project.
They are reproduced every ~5 minutes of running a script that constantly launches and terminates my program.
After some deep debugging I have found several bugs, and this commit fixes the issue for me(~1 hour run - no issues).
Fixes:
1. poll_*_ready called before processing
2. dbus_connection_flush called to ensure that all messages was sent
3. A way of waking up the IOResource implemented, so it can call flush
4. Mutex is held until handler for reply is created.

### DIFF
--- a/dbus-tokio/src/connection.rs
+++ b/dbus-tokio/src/connection.rs
@@ -100,15 +100,25 @@ impl<C: AsRef<Channel> + Process> IOResource<C> {
         // Make a promise that we process all events recieved before this moment
         // If new event arrives after calls to poll_*_ready, tokio will wake us up.
         let read_ready = is_poll_ready(watch_reg.poll_read_ready(ctx)?);
-        let write_ready = is_poll_ready(watch_reg.poll_write_ready(ctx)?);
         let send_ready = is_poll_ready(waker_reg.poll_read_ready(ctx)?);
 
-        if read_ready || write_ready || send_ready {
+        // If we were woken up by write ready - reset it
+        let _ = watch_reg.take_write_ready()?;
+
+        if read_ready || send_ready {
             loop {
                 c.read_write(Some(Default::default())).map_err(|_| Error::new_failed("Read/write failed"))?;
                 self.connection.process_all();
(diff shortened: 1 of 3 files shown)


diwic:
Hi! Thanks for wanting to help out! 

Let's go through them one by one:

 > poll_*_ready called before processing

After some thinking, I think this is a remain from tokio 0.1, which was level triggered. Now that we're edge triggered, this change should be correct.

 > dbus_connection_flush called to ensure that all messages was sent

This function "Blocks until the outgoing message queue is empty", according to documentation. We should not block the event loop. We need some other way to solve this.

 > A way of waking up the IOResource implemented, so it can call flush

And since we should not call flush, I'm not sure where that leaves us with this one.

 > Mutex is held until handler for reply is created.

Correct. Good catch! Just remove the "TODO" comment, as I don't think we should use some other Mutex here.

Snarpix:
Ok, new plan.
If DBus for some reason has unsend messages - we check if fd is writable or write would block.
If fd is writable - we try again.
I have tested it - seems to work.
Previous patch with c.flush() line removed - fails in 10 minutes.
New patch - 1 hour no problems.

diwic:
Aha, nice find of `dbus_connection_has_messages_to_send`! 

 1) Can we also use this in the `channel.send` task? Because libdbus makes an attempt to send the message, which should succeed in most cases, and then we should not need to wake up the other task.

 2) You skip the entire loop if `!(read_ready || send_ready)`, should we not go into the loop in case we were woken up by tokio because we can write more to the socket? Something looks weird here.

Snarpix:
1. Yes, I think it should work, added.
2. Yeah, good catch, fixed.

I also added propper error handling.

diwic:
I'm not completely sure that the write handling is correct, but it's likely better than before, so I think it's good enough for me at this point, so I merged it. 🙂 

Thanks for contributing!

PR #260, opened 2020-05-03
Fix incliding the repo as a Cargo dependency.

deifactor:
cargo tries to resolve all the individual crates, and the dbus-futures Cargo.toml didn't get its path updated when it moved into the attic, so it'd look for the dbus crate in attic/dbus.

### DIFF
--- a/attic/dbus-futures/Cargo.toml
+++ b/attic/dbus-futures/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 futures-preview = { version = "0.3.0-alpha.16" }
-dbus = { path = "../dbus" }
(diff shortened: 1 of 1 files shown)


diwic:
Sure, but I was hoping cargo wouldn't try that in the first place. What command are you running that actually makes cargo consider `dbus-futures` to be a crate? It's not in the workspace.

deifactor:
Just doing `dbus = { git = "https://github.com/diwic/dbus-rs" }`. It looks like it looks at every Cargo.toml trying to find one named `dbus`, and it breaks when parsing the one in `dbus-futures`? I dunno.

Agreed that Cargo shouldn't try it, though. Maybe I'll file a bug.

PR #261, opened 2020-05-07
Make dbus dependency optional in dbus-codegen

Xaeroxe:
I expect this might be a controversial patch. Therefore I want to state right from the get go: No hard feelings if you just want to close this PR.

When cross compiling a project using dbus it feels silly to have to depend on host architecture dbus development libraries, and yet that's exactly what a project of mine has to to do. I spotted some ways to remove this dependency from dbus-codegen, but it comes at a few costs.

1. ArgType information is duplicated into dbus-codegen. I don't expect this to present a significant maintenance burden, but I'm ill informed on the topic.
2. ~~Usage as a binary is completely eliminated. Because cargo doesn't have [something like this](https://github.com/rust-lang/rfcs/pull/2887) I'm unable to specify the dbus dependency for just the binary crate usage.~~ One alternative would be stripping out the ability to obtain the dbus introspection xml from a running dbus service, and leaving the binary usage. EDIT: Did this via a cargo feature ~~I'm not sure how much the project gains from that, but then again I'm not a binary user.~~

Yet another option available to us is to separate out the library and binary into separate crates, which would enable us to declare a dbus dependency for just the binary.

Anyways, I'm curious how the project maintainers feel about these options. Thanks for reviewing my PR!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,4 @@ script:
   - cargo update
   - cargo check --all
   - cargo test --all -- --nocapture --color always
+  - cd dbus-codegen && cargo test --all --no-default-features -- --nocapture --color always
--- a/dbus-codegen/Cargo.toml
+++ b/dbus-codegen/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dbus-codegen"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["David Henningsson <diwic@ubuntu.com>"]
 description = "Binary crate to generate Rust code from XML introspection data"
 license = "Apache-2.0/MIT"
@@ -9,6 +9,9 @@ repository = "https://github.com/diwic/dbus-rs"
 keywords = ["D-Bus", "DBus"]
 readme = "README.md"
 edition = "2018"
+
+[features]
+default = ["dbus"]
 
 [lib]
 path = "src/lib.rs"
@@ -19,8 +22,13 @@ path = "src/main.rs"
 
 [dependencies]
 xml-rs = "0.3"
-dbus = { path = "../dbus", version = "0.8" }
-clap = "2.20"
+dbus = { path = "../dbus", version = "0.8", optional = true }
+structopt = "0.3"
+strum = "0.18"
+strum_macros = "0.18"
+
+[dev-dependencies]
+dbus = { path = "../dbus", version = "0.8" } # Same as above, but not optional
 
 [badges]
 is-it-maintained-open-issues = { repository = "diwic/dbus-rs" }
--- a/dbus-codegen/src/generate.rs
+++ b/dbus-codegen/src/generate.rs
@@ -1001,7 +1001,7 @@ static FROM_DBUS: &'static str = r#"
     #[cfg(feature = "dbus")]
     #[test]
     fn argtype_is_complete() {
-        let mut foreign = dbus::arg::ArgType::iter().map(|e| e as u8).collect::<Vec<_>>();
+        let mut foreign = dbus::arg::ArgType::all().iter().map(|e| *e as u8).collect::<Vec<_>>();
         let mut internal = super::ArgType::iter().map(|e| e as u8).collect::<Vec<_>>();
         foreign.sort_unstable();
         internal.sort_unstable();
--- a/dbus-codegen/src/generate.rs
+++ b/dbus-codegen/src/generate.rs
@@ -1,8 +1,8 @@
-
-use std::{io, error, iter};
 use std::collections::HashSet;
-use dbus::arg::ArgType;
+use std::{error, io, iter};
 use xml;
+
+use strum_macros::{Display, EnumIter, EnumString, EnumVariantNames};
 
 fn find_attr<'a>(a: &'a Vec<xml::attribute::OwnedAttribute>, n: &str) -> Result<&'a str, Box<dyn error::Error>> {
     a.into_iter().find(|q| q.name.local_name == n).map(|f| &*f.value).ok_or_else(|| format!("attribute not found: {:?}", n).into())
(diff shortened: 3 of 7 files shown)


diwic:
Hi and thanks for the PR!

Indeed it would be a hard blow just to wipe the binary away! But I think I understand your point as well: transforming xml introspection to code should be a dependency free process.

I was wondering whether it would be possible to do this as a feature? Basically if we had a `dbus` feature enabled by default, that you could disable in order to build without dependencies.

 > One alternative would be stripping out the ability to obtain the dbus introspection xml from a running dbus service, and leaving the binary usage.

Yes, I would prefer the binary still being built, but without the switches that allows for a connection to dbus.

Xaeroxe:
@diwic That's a great idea! I went ahead and wrote it and force pushed that to this PR. Additionally I refactored the program to use `structopt` as it was easier to implement your idea using that crate.

diwic:
Ok, I have now looked at your patch:

 * What I'm hesitant to is the addition of new dependencies (structopts, strum, strum-macros). 
Especially in dbus I try to keep dependencies low. I like your ambition to test that the argtypes match, but better loop through something like `0..255.map(|x| dbus::ArgType::from_c_int(x)` in that case. Or add an `ALL_ARG_TYPES` array manually in `dbus` if your prefer, and then copy-paste that into `dbus-codegen`. It's just not worth adding additional dependencies to make a test case in `dbus-codegen` happy. ArgType in dbus-codegen could probably be refactored, remove the things that are unused, change `from_i32` to `from_char` etc.

 * In dbus-codegen, I found the `structopt` / `strum` code less readable than the current code, mostly because I find proc-macros in general quite unreadable, as it requires very deep knowledge of what that macro does. I also find implicit matches `Some("fn") => MethodType::Fn` makes it very clear what becomes what and also allows for future things like `Some("rusts") => Yeast::Mold` where names don't match. This might be mostly a matter of taste, but then I'm the one who needs to maintain it...

 * Also, maybe we can move the things requiring `feature(dbus)` into a separate module? There would still be a `feature(dbus)` or two in `main.rs` but not as many as now.

Those were the main things I saw when looking at your patch set. Force-push is good, keep doing that.

Xaeroxe:
Regarding your first point, that's fair yeah. I was hesitant to add the dependencies to dbus itself, but I figured the benefit of automatically verifying the `ALL_ARG_TYPES` const was correct in a reliable automated manner outweighed the down side. However, since as you've mentioned that can be accomplished without adding a dependency, I'm inclined to remove that.

As for the second point, I'd like to start off with a philosophical question, both `clap` and `structopt` have excellent documentation, so if you don't need to read the source for either the function, or the proc macro, aren't a function and proc macro essentially equivalent in regards to "deciphering what this is doing"? If analyzing only the characters presented, what's the difference between 

```rust
    /// Type of server method
    #[structopt(
        short,
        long,
        value_name = "Fn",
        default_value = "Fn",
        parse(from_str = lowercase_to_enum),
        possible_values = &MethodType::VARIANTS,
        case_insensitive = true,
    )]
    methodtype: MethodType,

...

fn lowercase_to_enum<T: IntoEnumIterator + Display>(src: &str) -> T {
    assert!({
        let mut lower_names = T::iter().map(|s| s.to_string().to_lowercase()).collect::<Vec<_>>();
        lower_names.sort_unstable();
        lower_names.windows(2).all(|w| w[0] != w[1])
    });
    let src = src.to_lowercase();
    for variant in T::iter() {
        if src == variant.to_string().to_lowercase() {
            return variant;
        }
    }
    unreachable!("possible_values from clap will guarantee we never arrive here.")
}
```

and 

```rust
.arg(clap::Arg::with_name("methodtype")
.short("m")
.long("methodtype")
.takes_value(true)
value_name("Fn")
.help("Type of server method; valid values are: 'Fn', 'FnMut', 'Sync', 'Generic', and 'None'. Defaults to 'Fn'."))

...

    let mtype = matches.value_of("methodtype").map(|s| s.to_lowercase());
    let (mtype, crhandler) = match mtype.as_ref().map(|s| &**s) {
        None | Some("fn") => (Some("MTFn"), None),
        Some("fnmut") => (Some("MTFnMut"), None),
        Some("sync") => (Some("MTSync"), None),
        Some("generic") => (Some("MethodType"), None),
        Some("par") => (None, Some("Par")),
        Some("none") => (None, None),
        _ => panic!("Invalid methodtype specified"),
    };
```

There are also benefits to this approach. I sneaked in a bug fix by eliminating the possibility for it to occur. The help for `methodtype` originally reads `Type of server method; valid values are: 'Fn', 'FnMut', 'Sync', 'Generic', and 'None'. Defaults to 'Fn'.` Except it missed a valid value. `Par`. This is natural and to be expected, because documentation goes out of date, and incorrect strings don't present compiler errors. The help message now reads `Type of server method [default: Fn]  [possible values: Fn, FnMut, Sync, Generic, Par, None]` and will always be informed by the `enum` definition, because we automated the generation of that help string. Additionally, there's less busy work that needs to be done now, because the string to match against for each enum is now automatically generated. I also went to great lengths while building this to ensure it's 100% backwards compatible with the prior argument structure, and I tested it against a few test commands.

Other considerations for this patch include

1. [`clap` incorporated `structopt` into version 3.0 as `clap-derive`.](https://github.com/clap-rs/clap/blob/master/FAQ.md#how-does-clap-compare-to-structopt) To me, this indicates a shift from the `clap` author(s) that `structopt` is the intended front end for `clap`. [Indeed, the first example `clap` provides, now leans on proc macros.](https://github.com/clap-rs/clap#quick-example)
2. [`structopt` exposes all `clap` methods via the raw attributes](https://docs.rs/structopt/0.3.14/structopt/#raw-methods), in fact this patch uses many of them.

Finally, even if you do want the string argument and the enum variant to have different names you can still do that by providing an alternate `parse` method, as well as an alternate `possible_values`.

So, with all that in mind, do you still want me to back out my `structopt` changes? If I did, then I would also make sure to update the doc string for `methodtype` so as to preserve my bug fix.

Xaeroxe:
So I went to implement `0..255.map(|x| dbus::ArgType::from_c_int(x)` and I ran into a few problems.

`dbus::ArgType::from_i32` is defined in terms of `ALL_ARG_TYPES` so I'd be checking the constant against itself, and that defeats the purpose. So in order to remove that dependency I'd have to alter the unit test in such a way that `ALL_ARG_TYPES` can't be automatically verified, and that defeats the point of the unit test.

diwic:
I've pushed a new function `dbus::arg::ArgType::all()` that should aid verification in dbus-codegen.

diwic:
> aren't a function and proc macro essentially equivalent in regards to "deciphering what this is doing"?

No. A proc macro can do anything to your struct, e g add new methods, fields, change field names or whatever on that struct - very undecipherable (without the documentation). Functions have to be imported with `use` statements (or use qualified paths). 

That said, I guess I can learn to live with `structopt` or `clap-derive` as well, if this is the future of clap anyway. Was just a bit overwhelmed with all the changes at once. 

 > I sneaked in a bug fix by eliminating the possibility for it to occur. The help for methodtype originally reads Type of server method; valid values are: 'Fn', 'FnMut', 'Sync', 'Generic', and 'None'. Defaults to 'Fn'. Except it missed a valid value. Par. 

Actually, `Par` is just some internal test code, and I'm going to remove it when I get some time. If anything, it's actually better that it isn't currently visible, but this is not a big deal either. I can see the use in it being automatically handled too.

    assert!({
        let mut lower_names = T::iter().map(|s| s.to_string().to_lowercase()).collect::<Vec<_>>();
        lower_names.sort_unstable();
        lower_names.windows(2).all(|w| w[0] != w[1])
    });

I think this was part of the confusion actually. It belongs to code that runs when you call `cargo test`, not running as part of argument parsing.

Xaeroxe:
I guess I got a bit excited in my quest to automatically verify invariants and now this PR is conflated with other things. I'll remove some of the heavier unit tests and get rid of the assert. The assert is trying to catch a case we can already pretty confidently say will never happen.

Xaeroxe:
[I think this commit resolves the outstanding concerns.](https://github.com/diwic/dbus-rs/pull/261/commits/1e523b4882d7de98dd69cd2fe1e7fdc2e3eda9dc)

Xaeroxe:
[Here's another option we can try. If you don't like it I'll back out the commit.](https://github.com/diwic/dbus-rs/pull/261/commits/547dbd4d77ad402a7ca8bbcfd340cfc2f8688b21)

diwic:
Hi,

So given this some more thoughts and possibilities to refactor code, I realized it was easier for me to write the code directly, than to communicate my wishes to you. 

So the feature you were looking for has now implemented, but perhaps not the way you preferred. Hope that's okay with you.

PR #262, opened 2020-05-13
Add missing PropertiesChanged signal to introspection

paulvt:
This signal is part of the `org.freedesktop.DBus.Properties` interface which is also available as the
`dbus::*::stdintf::org_freedesktop_dbus::PropertiesPropertiesChanged` signal struct. This interface is added by default for object paths, but the `PropertiesChanged`signal is not part of the introspection data. This commit adds this.


diwic:
Hmm, weird that this went undetected for so long. Thanks for fixing!

paulvt:
Is there any play to have an 0.8.3 release soon? The master branch has been accumulating many fixes.

diwic:
@paulvt Sure! Done now, enjoy 😀

PR #268, opened 2020-05-28
codegen: impl generated trait for all BlockingSender instead of only blocking::Connection

iovxw:
Just like [`stdintf::org_freedesktop_dbus::Properties`](https://docs.rs/dbus/0.8.3/dbus/blocking/stdintf/org_freedesktop_dbus/trait.Properties.html#implementors)

This won't break compatibility, since Deref is a trait with only associated types (correct me if I'm wrong)


diwic:
Thanks!

PR #270, opened 2020-07-15
Only depend on futures-{util,channel}

coolreader18:
This cuts out a lot of unused transitive dependencies (`futures-{executor,io,task}`).


diwic:
Okay, let's try it and hope that no one complains.

Issue #166, opened 2019-02-11
dbus-codegen: Option to generate async code


zeenix:
`dbus-codgen` should be able to generate async code that makes use of `dbus-tokio` crate. I plan to work on this (or get a GSoC student to) as requirement for https://gitlab.freedesktop.org/geoclue/geoclue/issues/103 but adding it here already in case someone wants to beat me to it. :)

diwic:
So I pushed a commit earlier today for dbus-futures, so begin by checking git master :-)

Although that is the initial client-side work only, not server part.

What I'm thinking now is deprecate dbus-tokio when futures 0.3 / async / await / etc becomes stable, and have dbus-futures support Tokio as an event loop backend.
Do you need an intermediate solution until that stuff is stable?

zeenix:
> So I pushed a commit earlier today for dbus-futures, so begin by checking git master :-)

Hehe, sure. :) Thanks.

> Although that is the initial client-side work only, not server part.

Right and geoclue needs the server side.

> What I'm thinking now is deprecate dbus-tokio when futures 0.3 / async / await / etc becomes stable, and have dbus-futures support Tokio as an event loop backend.

Sounds good.


>  Do you need an intermediate solution until that stuff is stable?

I'm neither sure about my own timeline on this, nor that of future 0.3/async/await becoming stable. :( What I do know is that I plan to start the basic implementation (dummy one at least) before summer at latest, so I know how the new code architecture would be like and hence can guide the student(s) during the GSoC.

diwic:
> I'm neither sure about my own timeline on this, nor that of future 0.3/async/await becoming stable. :( What I do know is that I plan to start the basic implementation (dummy one at least) before summer at latest, so I know how the new code architecture would be like and hence can guide the student(s) during the GSoC.

I think it kind of depends on how future proof you want your code to be as well.

It seems like async/await/futures 0.3 is becoming stable during 2019, although you don't know for sure until it's there. But if it's semi-stable, then you can work with the nightly compiler until it's stable. 
Whereas Tokio + Futures 0.1 just might be deprecated soon after Futures 0.3 is stable. And then it might make less sense to develop new code for Futures 0.1. Anyhow I've put my eggs in the Futures 0.3 basket for now.

zeenix:
> I think it kind of depends on how future proof you want your code to be as well.

[...]

Issue #244, opened 2020-03-21
dbus-rs could use some fresh blood


diwic:
My motivation for maintaining dbus-rs has slowly declining over the past year or so. The two main reasons are:

1) I don't have a project that actually uses dbus-rs. When I started this binding five years ago, I had grand plans of software I wanted to build on top of these bindings. Life however had other priorities and those grand plans never happened.

2) One of the hardest part when writing Rust is coming up with the best design. Especially so when it comes to dispatching events, which is not Rust's best fit, but an important part of dbus-rs nonetheless. Having people to discuss these things with would help a lot here.

Having a co-maintainer would help a lot both with the over all quality of these bindings, as well as making coding more fun again. Or if this is not an option for you; consider if you can contribute to this library in another way.

Finally - Rust's ecosystem needs production quality bindings to D-Bus and so far, this project has provided them. I'd like it to stay that way.

snprajwal:
@diwic I don't know how far I qualify to be a co-maintainer of this project, but I'd love to help out in any way I can. I'd be happy to review PRs and test things in my free time. I would not want this project to stall or be retired, because it provides something none of the other current dbus implementations provide - synchronous calls without the overhead of an async runtime.

- `rustbus` requires writing too much boilerplate to get stuff working
- `zbus` pulls in async-std/tokio even if blocking connections are used. Furthermore, the blocking API is just a wrapper around the non-blocking async calls, which bloat the binary with all the state machine logic even if it's not being used in any meaningful way

It would be great to have this project alive and maintained, if only to provide a simple and practical alternative to `zbus` for smaller projects.

diwic:
@snprajwal Thanks!

So to sum up the state four years later:

 * The bindings are mature and do what most people need them to do
 * The project is alive and maintained in the sense that if there is a bug and people post PRs then I will review, merge and release as requested by the bug reporter
 * There are some larger issues (marked with "help wanted") that I'm not likely to do myself any time soon
 * We still bind to a C library by default, but cross-compiling is easier due to the `vendored` feature that builds the C code along with the Rust code.
 * Contributions are welcome (se readme.md in the root)

For you or anyone else wanting to become a co-maintainer; let me know your visions for the project, see we can agree on some common goals, then start working on those over time to show that you're in it for the long run. Reviewing and testing is welcome but for co-maintainership or even handover I think I'd want someone that drives the project forward.
Does that make sense?

snprajwal:
[...]

PR #276, opened 2020-08-30
codegen: reenable --futures flag

alsuren:
This is part of an effort to produce a non-blocking bluez client library, for use in https://github.com/alsuren/reading-xiaomi-temp/pull/27 . I would really rather not maintain my own bluez library manually, but I'm happy to maintain a set of bindings generated from the spec.

to test:
```
cargo install --git=https://github.com/alsuren/dbus-rs/ --branch=generate-futures dbus-codegen
# TODO: it would probably be enough to take anything that matches `/org/bluez/hci0/*` because they all seem to implement the same interface.
dbus-codegen-rust --system-bus --destination org.bluez --path /org/bluez/hci0/dev_A4_C1_38_1E_0A_E8 --client nonblock --futures > bluez-generated/mod.rs
```

I will take this pull request out of Draft once I have something that generates working client bindings. If there is a particular direction that you would like me to take this in (like folding the `futures` flag into the `ConnectionType` enum or something) then please say.

I wasn't planning to fix the server-side codegen as part of this PR (so it will only go part-way to addressing https://github.com/diwic/dbus-rs/issues/166).

[edit: it turns out that it's already possible to generate non-blocking client interfaces as long as you disable server method generation with `--methodtype none`. I'm going to verify that this generates usable code and then think about switching this PR around to delete all `--futures`-related code]

### DIFF
--- a/dbus-codegen/src/main.rs
+++ b/dbus-codegen/src/main.rs
@@ -46,8 +46,8 @@ Defaults to 'RefClosure'."))
              .help("Name of dbus crate, defaults to 'dbus'."))
         .arg(clap::Arg::with_name("skipprefix").short("i").long("skipprefix").takes_value(true).value_name("PREFIX")
              .help("If present, skips a specific prefix for interface names, e g 'org.freedesktop.DBus.'."))
-//        .arg(clap::Arg::with_name("futures").short("f").long("futures")
-//             .help("Generates code to use with futures 0.3 (experimental)"))
+       .arg(clap::Arg::with_name("futures").short("f").long("futures")
+            .help("Generates code to use with futures 0.3 (experimental)"))
         .arg(clap::Arg::with_name("client").short("c").long("client").takes_value(true).value_name("client")
              .help("Type of client connection. Valid values are: 'blocking', 'nonblock', 'ffidisp'."))
         .arg(clap::Arg::with_name("output").short("o").long("output").takes_value(true).value_name("FILE")
@@ -120,7 +120,7 @@ Defaults to 'RefClosure'."))
         skipprefix: matches.value_of("skipprefix").map(|x| x.into()),
         serveraccess: maccess,
         genericvariant: matches.is_present("genericvariant"),
-        futures: false,
+        futures: matches.is_present("futures"),
         connectiontype: client,
         crhandler: crhandler.map(|x| x.to_string()),
         interfaces,


alsuren:
Closing this PR. I will open separate PRs for the things that I noticed while making this PR.

Regarding #244: I am interested in helping out. Most of my OSS contributions are a bit drive-by these days, so I will start off with small PRs and see how it goes. I don't run desktop linux anymore, but I still mostly remember how dbus fits together, from my days on the telepathy team at Collabora a decade ago. I'm also happy to jump on a call and brainstorm or pair (Europe/London time).

Some notes about possible PRs to make once I have finished my bluetooth refactor in my own repo:

- [ ] Delete the code for the --futures flag (needs a semver-breaking change to the codegen library, as we remove a bool from a public struct).
- [ ] Make an example that explicitly shows how to pluck a HashMap<String, Variant> out of a Variant in a nice way (nested `a{sv}`s are quite a common use-case (see get_managed_objects(), below), so it would be good to have some more signposts along the way).
- [ ] run rustfmt over the codebase, to reduce the barrier of entry for contributors
- [ ] clear out the attic, and generally remove dead/commented-out code.
- [ ] Find a way to generate server-side async code to implement #166 (probably using async-trait or something?)
- [ ] Find a way to generate something like `impl OrgBluezDevice1` from the return value of `get_managed_objects()`. This way you could do something like this, and avoid a few layers of guessing dictionary keys variant types:
```rust
let bluez_root =
    dbus::nonblock::Proxy::new("org.bluez", "/", Duration::from_secs(30), conn.clone());
let proxy_wrapper = ManagedObjectsFakeProxy::new(
    bluez_root.get_managed_objects().await?
);
proxy.map_interface(
    "org.bluez.Device1",
    |fake_proxy_object| {
        // `address` could then be a String if the object has such a property, or TypeMismatchError if it does not,
        // without making any more round-trips, and without dealing with quite so many Variants
        let address = fake_proxy_object::address()?;
        ...
        Ok(())
    }
)
```

If there are any ideas that you would like me to focus my energy on first, please drag them to the top of the list and/or add a comment. Otherwise, I will pick things up as I feel the urge.

diwic:
Hi! You're of course very welcome to help out!

 > Make an example that explicitly shows how to pluck a HashMap<String, Variant> out of a Variant in a nice way (nested a{sv}s are quite a common use-case (see get_managed_objects(), below), so it would be good to have some more signposts along the way).

This one would be my favorite given that there have been more than one question about this over the years. If more helper methods would improve the situation, feel free to add them.

 > run rustfmt over the codebase, to reduce the barrier of entry for contributors

This one is a NAK. I've really tried to make friends with rustfmt, but failed. It slows me down considerably, and makes the code look more ugly.
If we're ever at the point where you take over the repo completely and commit to maintain it for the forseeable future, it would be your decision to make, but until then, it's a no.

 > Find a way to generate server-side async code to implement #166 (probably using async-trait or something?)

dbus-crossroads is the way forward as the async support is much better from the start: I would prefer if you made async code for dbus-crossroads rather than dbus::tree, or possibly both.

diwic:
@alsuren I pushed a commit that adds a `prop_cast` method. It's tiny, but I think this would be helpful both as it is, and as a helper method for your last bullet point. Have a look!

diwic:
@alsuren You might be interested in something from another contributor, see the new `--prop-newtype` option for codegen.

alsuren:
> @alsuren You might be interested in something from another contributor, see the new `--prop-newtype` option for codegen.

Yeah, my housemate wrote it 
😀. He continues to have a lot more energy for these things than I do (sorry for offering to help and then running out of steam).

If anyone is interested, this is how we're using it: https://github.com/alsuren/mijia-homie/pull/120. We have also released a higher level bluez-async crate, but it has been really valuable to have bluez-generated as a generated-code-only crate (I like to think of it like a -sys crate).

There might be some value in generating a bunch of crates that wrap standard desktop Linux services, and publishing them on crates.io. The async generated interfaces are really nice to work with once you have them, but it's not always obvious that this is even an option. We spent quite a long time fighting with the old single-threaded, blocking interface that was exposed by blurz, before I finally ripped it out and replaced it with bluez-generated. (I only knew that this was a good idea because of the time that I spent working on telepathy back in the day). I suspect that other people will be in a similar situation with other Linux interfaces, and having generated crates on crates.io would be enough to signpost them in the right direction. 

(Excuse any typos: I'm typing on my phone)

PR #277, opened 2020-09-01
Do not include the ManagedObjects implementor in the result

mulkieran:


mulkieran:
test_prop_handlers failed...I'll take a look at that also.

diwic:
Hi and thanks for the PR!

I'm okay with the functional change, but please adjust the test case (or write a new one) so that objectmanager is still tested - i e, there needs to be a subpath under `/test` that the object manager test can retrieve values from.

mulkieran:
> Hi and thanks for the PR!
> 
> I'm okay with the functional change, but please adjust the test case (or write a new one) so that objectmanager is still tested - i e, there needs to be a subpath under `/test` that the object manager test can retrieve values from.

Thanks! Will do.

mulkieran:
@diwic I think it is ready now.

diwic:
Thanks!

PR #280, opened 2020-09-03
Implement std::error::Error for MethodErr

ids1024:
This is generally a good practice for types that are used as errors with `Result`, and it seems easy enough to do here.


diwic:
Thanks! I'm kind of surprised that this wasn't already done. 🙂

PR #281, opened 2020-09-07
blocking: regenerate standard interfaces

lopsided98:
This PR is based on #217, but applied to the blocking interface. Regenerating these files allows them to be used with LocalConnection and SyncConnection now that #268 is merged.


diwic:
Thanks!

PR #284, opened 2020-09-21
Update monitor.rs

xkr47:
Update to support newer monitoring scheme using BecomeMonitor call; on failure fall back to old method for compatibility with older systems.


diwic:
Thanks!

PR #287, opened 2020-10-14
Add some additional methods to Message

ids1024:


diwic:
Hi and thanks for the PR!
Usually I just pull a PR that just adds methods, but this one had me wondering. Mostly because you're more likely to misuse these particular ones rather than use them, according to my gut feeling. What is your actual use case for these methods?

ids1024:
I'm trying to send a response to a method call in a separate thread (with the same `Arc<SyncConnection>`), and it can't use `.method_return()` or `.error()` since `Message` can't be cloned to send to the other thread.

I thought about implementing `Clone` for `Message`, but looking at the documentation for `dbus_message_copy`, that seems too problematic.

(Perhaps porting the code in question to use `dbus-tokio` would allow dealing this this sort of thing better...)

diwic:
Hmm...I'm not following, but this does not seem to be the right solution to your problem. `Send` is implemented for `Message`, but I don't understand why you need to create a clone to send (instead of sending the original)?

ids1024:
The code was using `dbus::tree`. The method handlers for that are passed just an `&Message`, then I was having it return an empty list of messages to send, but spawn a thread that will send the reply on the same connection.

I don't think there's technically anything wrong with that, but it's definitely a bit of an ugly hack. I've since started porting the old code to use `dbus-tokio` and `dbus-crossroads`, which should be cleaner.

diwic:
Given your reply I believe it would be okay to close this PR without merging it.

It could possibly be useful to have two methods, maybe `error_reply_from_serial(reply_serial, name, message_text) -> Message` and `reply_from_serial(reply_serial) -> Message` because it is easier to send and copy a reply_serial than a full message.

PR #288, opened 2020-10-16
Fix examples that aren't compiling

ids1024:


diwic:
Thanks! I should fix the CI to make sure it catches this type of errors...

PR #291, opened 2020-10-27
Use {}, not {:?} in implementation of Display for Error

ids1024:
Consider this example:

```rust
fn main() {
    println!("{}", dbus::Error::new_failed("Foo's bar"));
}
```

Before this change, this would output `"Foo\'s bar"`.

It now outputs `Foo's bar`.

Escaping the ' seems clearly undesired. Wrapping it all in parentheses is sometimes desirable, but then the caller should do it.


diwic:
Makes sense to me. Thanks!

Issue #293, opened 2020-11-12
Panic when using object manager on `/`


zevweiss:
When using the object manager interface on the path `/`, calling `GetManagedObjects` on it leads to a panic (assuming there's at least one subpath).

Example program:

```rust
use std::{
	error::Error,
	sync::{Arc, Mutex},
};
use dbus::blocking::Connection;
use dbus_crossroads::Crossroads;

fn main() -> Result<(), Box<dyn Error>> {
	let bus = Connection::new_session()?;
	bus.request_name("foo.crashdemo", false, false, false)?;
	let mut cr = Crossroads::new();
	cr.set_object_manager_support(Some(Arc::new(Mutex::new(vec!()))));
	cr.insert("/", &[cr.object_manager()], ());
	cr.insert("/foo", &[], ());
	cr.serve(&bus)?;
	unreachable!();
}
```

With that running, executing `busctl --user call foo.crashdemo / org.freedesktop.DBus.ObjectManager GetManagedObjects` results in:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "Object path was not valid: \'//foo\'"', $HOME/.cargo/registry/src/github.com-1ecc6299db9ec823/dbus-0.9.0/src/strings.rs:187:1
```
From what I can tell, this appears to be due to [this line](https://github.com/diwic/dbus-rs/blob/b510fbbc2f598a5a2e00b5300b1844c6b15c0868/dbus-crossroads/src/stdimpl.rs#L348) unconditionally appending a `/` separator before the child path, which in the special case of the parent path being `/` leads to a leading double-slash, which `dbus_validate_path()` rejects.

PR #294, opened 2020-11-12
Fix edge case with object manager on '/'.

zevweiss:
In the special case of the object manager interface being applied on
'/', we need to skip appending the separator before the child path,
since it leads to a double-slash and hence an invalid path, which leads
to panicking down the road.

Fixes: #293

### DIFF
--- a/dbus-crossroads/src/stdimpl.rs
+++ b/dbus-crossroads/src/stdimpl.rs
@@ -345,7 +345,9 @@ fn get_managed_objects(mut ctx: Context, cr: &mut Crossroads, _: ()) -> Option<C
     let children: Vec<dbus::Path<'static>> =
         cr.get_children(ctx.path()).into_iter().map(|child_path| {
             let mut x = String::from(&**parent);
-            x.push_str("/");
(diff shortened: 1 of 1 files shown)


zevweiss:
I _think_ this is an appropriate fix, though I'm far from a dbus expert and may be missing something.

As for the check being `!x.ends_with('/')` instead of just `x != "/"`, I figure it's a matter of:

 - a simpler, faster check that should catch (what I think is) the only valid case where it's necessary

vs.

 - a slightly more expensive check that in the hopefully-unlikely scenario of somehow having an already-invalid parent path with a trailing slash avoids making it any _more_ invalid.

I'm not strongly opinionated and semi-arbitrarily went with the latter; if you'd prefer the former I'm happy to do it that way instead.

diwic:
Hi and thanks for the PR! As for `ends_with` vs `!=` any path, the D-Bus spec says:

    A trailing '/' character is not allowed unless the path is the root path (a single '/' character). 

So essentially it does not matter. If you like, also submit the test case you had in the issue as a test case, that would be nice!

PR #296, opened 2020-11-12
Add object_manager_root() test.

zevweiss:
This exercises the problem in #293.

[As suggested [here](https://github.com/diwic/dbus-rs/pull/294#issuecomment-726140143).]


diwic:
Thanks!

PR #297, opened 2020-11-13
Allow errors to be passed as reply in AConnection

Snarpix:
This fixes the issue, that in old AConnection interface error replies to method calls are ignored.


diwic:
Thanks!

PR #301, opened 2020-12-29
Tokio 1.0

### LINKED ISSUE #289: tokio 1.0 support
New version of tokio is released
current version of tokio is 0.2 [Cargo.toml](https://github.com/diwic/dbus-rs/blob/master/dbus-tokio/Cargo.toml)

Better upgrade, to avoid downloading two versions of tokio in projects

coolreader18:
Fixes #289 

Currently it's compiling and I think it's roughly correct, but it's timing out while waiting for a response, even though the last call to poll_internal comes because read is ready.

### DIFF
--- a/dbus-tokio/Cargo.toml
+++ b/dbus-tokio/Cargo.toml
@@ -14,14 +14,13 @@ readme = "README.md"
 
 [dependencies]
 dbus = { path = "../dbus", version = "0.9", features=["futures"] }
-mio = "0.6.21"
 libc = "0.2.69"
-tokio = {version = "0.2.4", features=["time", "macros", "io-driver", "rt-core", "rt-util", "rt-threaded", "stream"]}
+tokio = {version = "1.0", features=["time", "net"]}
 dbus-crossroads = { path = "../dbus-crossroads", optional = true, version = "0.3" }
 
 [dev-dependencies]
 futures = "0.3.1"
-tokio = {version = "0.2.4", features=["time", "macros", "io-driver", "rt-core", "rt-util", "rt-threaded", "stream", "blocking"]}
+tokio = {version = "1.0", features = ["macros", "rt-multi-thread"]}
 dbus-tree = {path = "../dbus-tree", version="0.9"}
 
 [badges]
@@ -30,7 +29,7 @@ is-it-maintained-issue-resolution = { repository = "diwic/dbus-rs" }
 travis-ci = { repository = "diwic/dbus-rs" }
 
 [[example]]
-name = "tokio02_server_cr"
+name = "tokio_server_cr"
 required-features = ["dbus-crossroads"]
 
 [[example]]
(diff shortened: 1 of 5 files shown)


coolreader18:
Weird, it looks like `IOResource::poll` never gets called, even though if I add a manual wake() call after send() in send_with_reply() it seems that poll_read_ready is Ready.

coolreader18:
Yeah, it looks like it can receive one message, and then poll_read_ready never wakes it up again.

coolreader18:
Figured it out, I ran into the same issue as https://github.com/tokio-rs/tokio/issues/3208#issuecomment-744534414

diwic:
Given your above fix, what's the current status of the work currently? Stuck somewhere? Ready for merging?

coolreader18:
Oh, yeah, this is pretty much ready to merge if it passes review.

dbus-tokio/Cargo.toml
```diff
@@ -14,14 +14,14 @@ readme = "README.md"
 
 [dependencies]
 dbus = { path = "../dbus", version = "0.9", features=["futures"] }
-mio = "0.6.21"
 libc = "0.2.69"
-tokio = {version = "0.2.4", features=["time", "macros", "io-driver", "rt-core", "rt-util", "rt-threaded", "stream"]}
+tokio = {version = "1.0", features=["time", "net"]}
+parking_lot = "0.11"
```
diwic, line 19: Unless there is a strong reason not to use std's Mutex, I would prefer to avoid additional dependencies.

dbus-tokio/src/connection.rs
```diff
@@ -69,55 +72,87 @@ enum IOResourceRegistration {
...
-    match i {
-        task::Poll::Ready(r) => !r.is_empty(),
-        task::Poll::Pending => false,
+/// An error that can occur in the dbus-tokio reactor
+#[derive(Debug)]
+pub struct IOResourceError {
+    inner: ResourceErrorKind
+}
+#[derive(Debug)]
+enum ResourceErrorKind {
```
diwic, line 85: How about we make the enum public and skip "IOResourceError"? Potentially with an added non_exhaustive attribute.

diwic (approved):
Ok! So I admit not knowing the inner details of Tokio and trusting that you do the right thing (or that other people will report bugs if you don't...). 
So I think this is okay for merging given the two comments above are fixed/commented + that you squash all patches into a single commit. 

Good work!

dbus-tokio/src/connection.rs
diwic: Also add `derive(Clone)` btw

coolreader18: Alright, wasn't sure whether I should add more public API surface area, since `Box<dyn Error>` is pretty anonymous, but I guess it makes sense that a dbus IO reactor would fail with either an IO or a dbus error :)

coolreader18: Oh, neither of the inner error types implement Clone.

diwic:
And it's in! Thanks!

Issue #251, opened 2020-04-10
Support: yet another variant extract question


Moret84:
Hello @diwic and thanks for your work on this crate.

I am new to rust and I am trying to extract a value from a variant, but the inner content seems unusal.

Here is the .0 debug format output:

``` rust
InternalDict {
    data: [
        (
            "0000fe95-0000-1000-8000-00805f9b34fb",
            Variant(
                [
                    80,
                    32,
                    170,
                    1,
                    85,
                    125,
                    189,
                    219,
                    168,
                    101,
                    76,
                    13,
                    16,
                    4,
                    229,
                    0,
                    109,
                    1,
                ],
            ),
        ),
        (
            "0000ffff-0000-1000-8000-00805f9b34fb",
            Variant(
                [
                    192,
                    75,
                    80,
                    56,
                    236,
                    7,
                ],
            ),
        ),
    ],
    outer_sig: Signature(
        "a{sv}",
    ),
}
```

I can't find how to get the inner data.
I have tried to cast it to a `HashMap<String, Vec<u8>>`, to a `HashMap<String, Variant<Box<dyn RefArg>>>` too, but without success.

This field come from the Bluez Device API. It is called ServiceData and the documentation describe it the following : "Keys are the UUIDs in string format followed by its byte array value."
https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/device-api.txt

I am sorry but I don't know where to ask . Could you help me ?

Thanks

diwic:
Yes, what you're looking at is an implementation detail which I added to fix another bug a while ago...

If what you got was a `Box<RefArg>` then you cannot directly cast the dict to something else. You need to use `as_iter()` instead, something like:

    let x = myprop.as_iter()?;
    while let Some(key) = x.next() {
       println!("{}", key.as_str()?);
       let value = x.next()?;
       let inner_value = x.as_iter()?.next()?; // descend into the variant
       let v: &Vec<u8> = arg::cast(&*inner_value)?;
       println!("{:?}", v);
    }

This is just straight out of my head, so not sure everything is correct above.

As an option, you might be able to get a `Variant<HashMap<String, Variant<Vec<u8>>>>` directly from the start. I can't say if this is easier or not as I have not seen your code.

Moret84:
First many thanks for your response.

> As an option, you might be able to get a `Variant<HashMap<String, Variant<Vec<u8>>>>` directly from the start. I can't say if this is easier or not as I have not seen your code.

[...]

PR #304, opened 2021-01-01
codegen: Generate struct to parse properties from HashMap.

qwandor:
This generates a typed wrapper around `prop_cast` for all gettable properties defined on an interface. It is useful both with `Properties::get_all` and with the interface list returned by the `ObjectManager`.

### DIFF
--- a/dbus-codegen-tests/build.rs
+++ b/dbus-codegen-tests/build.rs
@@ -177,6 +177,7 @@ fn main() {
     generate_code(POLICYKIT_XML, &g, "policykit_asref_mtsync.rs");
 
     g.methodtype = None;
+    g.propnewtype = true;
     generate_code(POLICYKIT_XML, &g, "policykit_client.rs");
 
 }
--- a/dbus-codegen-tests/tests/test_parse_properties.rs
+++ b/dbus-codegen-tests/tests/test_parse_properties.rs
@@ -1,0 +1,31 @@
+#[allow(dead_code)]
+#[deny(trivial_casts)]
+mod policykit_client;
+
+use dbus::arg::{RefArg, Variant};
+use policykit_client::OrgFreedesktopPolicyKit1AuthorityProperties;
+use std::collections::HashMap;
+
+#[test]
+fn test_parse_properties() {
+    let mut properties: HashMap<String, Variant<Box<dyn RefArg>>> = HashMap::new();
+    properties.insert("BackendFeatures".to_string(), Variant(Box::new(42u32)));
+    properties.insert(
+        "BackendName".to_string(),
+        Variant(Box::new("name".to_string())),
+    );
+    let mut interfaces = HashMap::new();
+    interfaces.insert(
+        "org.freedesktop.PolicyKit1.Authority".to_string(),
+        properties,
+    );
+
+    let authority_properties =
+        OrgFreedesktopPolicyKit1AuthorityProperties::from_interfaces(&interfaces).unwrap();
+    assert_eq!(authority_properties.backend_features(), Some(42));
+    assert_eq!(
+        authority_properties.backend_name().cloned(),
+        Some("name".to_string())
+    );
+    assert_eq!(authority_properties.backend_version(), None);
+}
--- a/dbus-codegen/src/generate.rs
+++ b/dbus-codegen/src/generate.rs
@@ -83,6 +83,8 @@ pub struct GenOpts {
     pub futures: bool,
     /// Type of connection, for client only
     pub connectiontype: ConnectionType,
+    /// Generates a struct wrapping PropMap to get properties from it with their expected types.
+    pub propnewtype: bool,
     /// interface filter. Only matching interface are generated, if non-empty.
     pub interfaces: Option<HashSet<String>>,
     /// The command line argument string. This will be inserted into generated source files.
@@ -93,7 +95,7 @@ impl ::std::default::Default for GenOpts {
     fn default() -> Self { GenOpts {
         dbuscrate: "dbus".into(), methodtype: Some("MTFn".into()), skipprefix: None,
         serveraccess: ServerAccess::RefClosure, genericvariant: false, futures: false,
-        crhandler: None, connectiontype: ConnectionType::Blocking,
+        crhandler: None, connectiontype: ConnectionType::Blocking, propnewtype: false,
         interfaces: None,
         command_line: String::new()
     }}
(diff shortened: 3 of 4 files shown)


diwic:
Hi and thanks for the PR!

Let's take a step back and discuss the general idea. I understand it like you got a hashmap of properties from somewhere, e g from `Properties::get_all` or `ObjectManagerInterfacesAdded` or something. Now you want to access those properties in an easy, type safe way. Correct?

Ok, so that makes sense. Next question is how to best implement this. It seems to me like it would be superfluous to create a struct `FooProperties` to but an `&HashMap` in, could you not make a `FooProperties` trait that you implement on the HashMap directly? That would enable you to make `set_xx` methods as well (that take `&mut self`).

Does that make sense to you?

qwandor:
Hi, thanks for the quick response!

Yes, that's what I'm trying to do. I did consider a trait implemented for `HashMap`. That would work, but I think it's a bit more messy and likely to lead to confusion because it doesn't let you distinguish a `HashMap` of properties for one interface from a `HashMap` of properties for another unless you can somehow cast from a `HashMap` to an `impl FooProperties`. For any given set of properties, we should know what interface it is for, and be able to use the appropriate wrapper struct to make that clear.

This is particularly relevant in the cases where `ObjectManagerInterfacesAdded` or `ObjectManager::get_managed_objects` returns a `HashMap` containing properties for all interfaces on an object. Suppose we have a `Foo` interface with a `length` property which is a `u32`, and a `Bar` interface with a `length` property which is an `f32`. Then we can do something like:

```rust
let interfaces_added: ObjectManagerInterfacesAdded = ...;
// interfaces_added.interfaces is a map from interface names to maps from property names to property values.
if let Some(foo) = FooProperties::from_interfaces(&interfaces_added.interfaces) {
  // foo has the concrete type FooProperties
  let length = foo.length().unwrap();
  // length has type u32
}
if let Some(bar) = BarProperties::from_interfaces(&interfaces_added.interfaces) {
  // bar has the concrete type BarProperties
  let length = bar.length().unwrap();
  // length has type f32
}
```

If we used a trait rather than a newtype, then as far as I can see `FooProperties::from_interfaces` would need to either return a `HashMap` (and lose the type information) or be a standalone function returning an `impl FooProperties` (because `impl` can't be used in trait method return types). Neither of these options seems as nice as my current approach of wrapping the reference in a struct.

Supporting typed setters is an interesting idea though. Where would that be useful?

diwic:
Hi!

I see. So, have you also considered making this just a struct, i e:

    struct FooProperties {
        pub length: u32,
        /* ... */
    }
    
    impl TryFrom<&HashMap<String, Variant<Box<RefArg>>>> for FooProperties {
        /* ... */
    }
    
That might be a slightly higher initial cost (performance wise), but it would even better for ergonomics, given that an API user 1) can then send this struct around as needed (because the struct has no reference to the original message) and 2) there is a single point of failure - when you create the struct - instead of having to error handle every property.

In extension, one could then also imagine an `all(&self) -> FooProperties` function for the `Foo` trait, that then would create this struct directly (i e without hashmaps and refargs in its implementation). But that's a future possibility.

qwandor:
Hello again. (-:

I tried that initially. Note that the fields of FooProperties in this case must all be typed like Option<u32>, because it's not guaranteed that every field will be provided. This is particularly true for `Properties::PropertiesChanged` messages where (at least in my experience so far) usually only one or two properties change at once, but even for `ObjectManager::InterfacesAdded` it can be that not all properties are provided. So I don't think the point about a single point of failure really stands, unfortunately.

The other question is what to do about properties of non-`Copy` types, like strings, arrays and maps. Either the FooProperties fields just store a reference (in which case we have the same lifetime issue as in my original proposal) or we clone everything. If we clone everything, then it seems a bit wasteful for some usecases. If I have an interface which has a large number of string properties with long values but I only care about a single integer property, then I end up having to clone all those strings just to get type-safe access to the one property I care about. I guess the cost of this is less than the cost of the original D-Bus message parsing, but I'm still concerned that the performance penalty might deter people from using the type-safe property interface.

From what I've seen of the rest of the `dbus` crate, it tends to use references quite a bit, so maybe this isn't so bad to do so here as well? I guess if we want the best of both worlds then I could take the same approach as `Path` and store a `Cow<HashMap<...>>` and add a `to_static` method, but it seems a bit overkill.

In case it's helpful, here's an example of these new structs in use: https://github.com/alsuren/mijia-homie/pull/115/files

Oh, and I'm still interested in your typed setters suggestion from your first comment. Do you have some idea in mind where that would be useful to have?

qwandor:
On further testing, I've found that this doesn't work when you have a property which is itself a map (or possibly also an array), because the value in the property map then ends up being a `Variant(InternalDict(...))` rather than a `Variant(HashMap(...))`, and so `prop_cast` fails to cast it to a `HashMap`.

I think the bug here is in how `RefArg` works rather than in the codegen, but I'm not sure what the correct fix is. In #251 you mentioned fixing `box_clone` to not destroy the `Vec`, and I guess the same might apply to a `HashMap`? How would this work? Alternatively, would it make sense for `as_any` on `InternalDict` convert itself to a `HashMap` to return?

diwic:
> Oh, and I'm still interested in your typed setters suggestion from your first comment. Do you have some idea in mind where that would be useful to have?
 
When I wrote it, it was more of a gut feeling than an actual use case. But:
 
One potential use case would be implementing PropertiesChanged, object manager etc on the server side. Not sure how often that would be used, given that we have dbus-crossroads and dbus-tree to help with that.
 
Also, trying to have an bird's eye view on how things work and fit together, I was brainstorming w r t splitting a D-Bus interface into three traits: methods, properties and signals. (Right now we have one trait with both methods and properties.) In which case the properties trait would need both getters and setters.

 > fixing box_clone to not destroy the Vec
 
I've tried, but I keep hitting limits w r t what you can do with object safe traits in Rust :-( But maybe I should have another go at it, it's a bit annoying...

 > the rest of the dbus crate, it tends to use references quite a bit
 
Not sure what this refers to. Most Rust code tend to use references quite a bit, it's a fundamental part of the language ;-)

Anyhow, the reason I'm considering other design choices is to figure if they could be useful in more use cases than yours. You seem to have already done that work (great!). But if we were to go with the newtype, maybe we could at least make it generic?

    struct FooProperties<T>(T);
    impl <T: AsRef<HashMap<..>>> FooProperties<T> {
       // getters here
    }

One could then later add setters (for `T: AsMut`) if someone else find a use case for it.

qwandor:
The generic `AsRef` idea looks promising, I'll try that and see if I can make setters work.

For the `InternalDict` issue I found, I've sent #306 which I think fixes it at least in the common cases. It may still be a problem for some complicated nested types. I don't think the issue was actually with `box_clone` after all.

qwandor:
I've had an attempt (but with `Borrow` rather than `AsRef`, as references don't automatically implement `AsRef`) and it kind of works, but is a bit less pleasant to use as the type parameter needs to be written out like `FooProperties<&HashMap<String, Variant<Box<dyn RefArg>>>>` each time `FooProperties` is used as a function parameter. I'll clean up the code and get it up here, but any thoughts on ergonomics?

qwandor:
Updated to make the newtype generic on `Borrow<HashMap<...>>`, and add setters for `MutBorrow<HashMap<...>>`. I have a slight preference for the original version, but if you think setters are useful then I'm happy with this too.

diwic:
Okay, so I guess it's down to me to make a decision then. So here's what I feel about it:

It's clearly useful for your use case, but I'm not sure how niche that use case is, or if the trait-impl-on-hashmap or prop-struct would work better in other use cases. So let's put it behind a flag for now, e g `--prop-newtype`. That also gives you the freedom to choose the original design or this one as you see fit. 

Btw, I've just added a type alias `PropMap = HashMap<String, Variant<Box<RefArg>>>` which might be useful if you want to save some characters.

qwandor:
Sounds good, done. I'll keep the `Borrow` version around in case we find a use for it later.

diwic:
Merged now. Thanks!

PR #306, opened 2021-01-02
Parse dictionaries of basic types to HashMap rather than InternalDict.

qwandor:
This allows `dbus::arg::cast` to work as expected for most `HashMap`s. Dictionaries with floating-point keys or array or struct values continue to use `InternalDict`.

I've added a test which fails without this change and passes with it.

This fixes the issue brought up in #251.

### DIFF
--- a/dbus/src/arg/array_impl.rs
+++ b/dbus/src/arg/array_impl.rs
@@ -489,29 +489,27 @@ where
     KF: FnMut(&mut Iter<'a>) -> Option<K>,
 {
     match value_type {
-        ArgType::Byte => get_dict_refarg::<K, u8, KF, _>(i, kf, Iter::get),
-        ArgType::Int16 => get_dict_refarg::<K, i16, KF, _>(i, kf, Iter::get),
-        ArgType::UInt16 => get_dict_refarg::<K, u16, KF, _>(i, kf, Iter::get),
-        ArgType::Int32 => get_dict_refarg::<K, i32, KF, _>(i, kf, Iter::get),
-        ArgType::UInt32 => get_dict_refarg::<K, u32, KF, _>(i, kf, Iter::get),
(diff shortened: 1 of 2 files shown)


diwic:
The reason I did not go with this approach myself, was the fear of code explosion. I mean, how many instances of `get_dict_refarg` will be monomorphized with this patch? 

That said, I too think the simple cases should be hashmaps. Could we just have one generic parameter and two subsections, `HashMap<K, Box<dyn RefArg>>` and `HashMap<K, Variant<Box<dyn RefArg>>>`? With the latter being a pragmatic reason that `String, Variant<Box<RefArg>>` being so common, and there would be a benefit from having the same representation.

qwandor:
Hmm, yeah, based on building a binary with and without this change, it appears that the impact on binary size of a release build on x86_64 is about 2.1 MB (and 25 MB for a debug build). That's more than I expected.

I don't quite understand your suggestion. What are you suggesting we do in the case that the dictionary value is not a `Variant`? As far as I can see `RefArg` is only implemented for `HashMap<K, V>` where `V` is `RefArg + Arg`, but `Arg` is not implemented for `Box<dyn RefArg>`. Is there some way around this?

diwic:
You're right in that my suggestion won't work. We can't make an `HashMap<K, Box<RefArg>>`. 
Would it make sense to special case either `String, Variant<Box<RefArg>>` or `K, Variant<Box<RefArg>>` into a HashMap and leave the rest as InternalDict?

qwandor:
Done. It doesn't seem like a great long-term solution, but it is enough for my usecases, so I'm happy to leave it at that. Do you want me to rebase this on top of the latest master?

qwandor:
Oh, and the binary size impact of this version is around 184 kB, which seems more reasonable.

PR #311, opened 2021-01-15
Apply the deprecated attribute if applicable interface/children are a…

thrykol:
…nnotated as such.  I'm unsure how to test that the attributes are added to the appropriate entity.  But a manual check confirmed the correct behavior.

### DIFF
--- a/dbus-codegen-tests/build.rs
+++ b/dbus-codegen-tests/build.rs
@@ -132,6 +132,28 @@ static POLICYKIT_XML: &'static str = r#"
 </node>
 "#;
 
+static DEPRECATED_XML: &'static str = r#"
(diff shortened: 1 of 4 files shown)


diwic:
Hi and thanks for the PR!

I would like to ask for one bigger change if you're okay with that - I'd like that all annotations are stored, not just the deprecated ones. So instead of 

    pub deprecate: Option<String>

I would like a

    pub annotations: HashMap<String, String>

...where all annotations are stored. I can then write more patches on top of that, that deal with other types of annotations. Does that make sense to you?

diwic:
Also when you believe your patch is ready to be merged, please squash it into one single commit. Thanks!

diwic:
Thanks!

PR #336, opened 2021-05-25
Remove extra futures dependencies

jeff-hiner:
Removes dependencies on futures-io, futures-macro (which has transitive deps on quote, syn, proc-macro2 and others), memchr, pin-project-lite, pin-utils, proc-macro-hack, proc-macro-nested, and slab. Should significantly speed up compilation when building with `futures` feature.


diwic:
Thanks!

PR #340, opened 2021-08-03
feat: support create connection from dbus channel

Sherlock-Holo:
Currently, the user only can create the async connection with `new*` function, with this PR, the user now can create its channel and convert it to the async connection.

For me, I want to connect the dbus with a custom unix path or TCP, but dbus-tokio doesn't allow me to do this job. WIth this PR, I can do it now :)

### DIFF
--- a/dbus-tokio/src/connection.rs
+++ b/dbus-tokio/src/connection.rs
@@ -214,9 +214,26 @@ fn make_timeout(timeout: Instant) -> pin::Pin<Box<dyn future::Future<Output=()> 
     Box::pin(t)
 }
 
-/// Generic connection creator, you might want to use e g `new_session_local`, `new_system_sync` etc for convenience.
-pub fn new<C: From<Channel> + NonblockReply>(b: BusType) -> Result<(IOResource<C>, Arc<C>), dbus::Error> {
-    let mut channel = Channel::get_private(b)?;
+/// Create a connection from channel, you may need to invoke `channel.register()?` to make sure the
+/// channel is usable.
+///
+/// # Example
+///
+/// ```rust
+/// use dbus::channel::Channel;
+/// use dbus_tokio::connection;
+///
+/// let mut channel = Channel::open_private("unix:path=/run/user/1000/bus").expect("open private channel failed");
+///
+/// channel.register().expect("register channel failed");
+///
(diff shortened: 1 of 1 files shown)


diwic:
Hi and thanks for the patch! It looks good to me but there is an error from the CI (see above). Could you have a look and correct the patch if necessary?

diwic:
Here's the error:

```
error[E0283]: type annotations needed for `(IOResource<C>, Arc<C>)`
   --> src/connection.rs:230:24
    |
11  | let (resource, conn) = connection::from_channel(channel).expect("create connection failed");
    |     ----------------   ^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `C` declared on the function `from_channel`
    |     |
    |     consider giving this pattern the explicit type `(IOResource<C>, Arc<C>)`, where the type parameter `C` is specified
    | 
   ::: /home/runner/work/dbus-rs/dbus-rs/dbus-tokio/src/connection.rs:236:24
    |
236 | pub fn from_channel<C: From<Channel> + NonblockReply>(mut channel: Channel) -> Result<(IOResource<C>, Arc<C>), dbus::Error> {
    |                        ------------- required by this bound in `from_channel`
    |
    = note: cannot satisfy `_: From<Channel>`
help: consider specifying the type argument in the function call
    |
11  | let (resource, conn) = connection::from_channel::<C>(channel).expect("create connection failed");
    |                                                ^^^^^

```

Sherlock-Holo:
@diwic the problem now is solved, `cargo check` report no errors

diwic:
@Sherlock-Holo So I merged it but I had to fix it up afterwards in https://github.com/diwic/dbus-rs/commit/6646184f9ec4453545c8dbfef2eece956aaaf330 - because the test suite should succeed even if the current logged in user does not have uid=1000. 
Let me know if you need a release.

Sherlock-Holo:
@diwic I will be very happy if you could bump a new version☺
and thanks for your commit to fix the problem that the test may failed if logined user uid is not 1000

diwic:
@Sherlock-Holo I'm all for making people very happy :-D 
0.7.4 of dbus-tokio relased now. Enjoy!

PR #347, opened 2021-09-06
Send Message to all matching MsgMatches

### LINKED ISSUE #346: If multiple match rules match a message, it only gets sent to the first one

qwandor:
If there are multiple `MsgMatch`es registered that match a message, it should be sent to all of them rather than just the first one. This requires a way to make a copy of the message, which I've done with `dbus_message_copy`. (It's not quite a clone, because it can fail.)

This fixes #346.

### DIFF
--- a/dbus/src/nonblock.rs
+++ b/dbus/src/nonblock.rs
@@ -178,13 +178,21 @@ impl Process for $c {
                 return;
             }
         }
-        let ff = self.filters_mut().remove_matching(&msg);
-        if let Some(mut ff) = ff {
-            if ff.2(msg, self) {
+        let matching_filters = self.filters_mut().remove_matching(&msg);
+        if matching_filters.is_empty() {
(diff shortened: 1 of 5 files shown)


dbus/src/nonblock.rs
```diff
@@ -178,13 +178,21 @@ impl Process for $c {
                 return;
             }
         }
-        let ff = self.filters_mut().remove_matching(&msg);
-        if let Some(mut ff) = ff {
-            if ff.2(msg, self) {
+        let matching_filters = self.filters_mut().remove_matching(&msg);
+        if matching_filters.is_empty() {
+            if let Some(reply) = crate::channel::default_reply(&msg) {
```
qwandor, line 183: I've kept this to preserve existing behaviour, but I don't understand what it's doing.

diwic:
This is not going to be merged as is, as per discussion in #346

PR #350, opened 2021-09-16
Fix signals not being send immediately

Snarpix:
I have fixed this issue for methods in #258, but not for signals. So here is the fix=)


diwic:
Thanks! Took a while to process that error condition, but I think the patch is correct so I merged it as is.

PR #351, opened 2021-09-23
Add option to send signals to all matching MsgMatches

### LINKED ISSUE #346: If multiple match rules match a message, it only gets sent to the first one

qwandor:
If there are multiple `MsgMatch`es registered that match a signal message, it should be sent to all of them rather than just the first one. This requires a way to make a copy of the message, which I've done with `dbus_message_copy`. (It's not quite a clone, because it can fail.) Following previous comments, I've made the new behaviour optional, and only for signals not other kinds of messages. Existing behaviour is maintained by default.

This fixes #346.

### DIFF
--- a/dbus/src/blocking.rs
+++ b/dbus/src/blocking.rs
@@ -4,7 +4,7 @@
 use crate::strings::{BusName, Path, Interface, Member};
 use crate::arg::{AppendAll, ReadAll, IterAppend};
 use crate::{channel, Error, Message};
-use crate::message::{MatchRule, SignalArgs};
+use crate::message::{MatchRule, SignalArgs, MessageType};
 use crate::channel::{Channel, BusType, Token};
 use std::{cell::RefCell, time::Duration, sync::Mutex};
 use crate::filters::Filters;
@@ -111,18 +111,21 @@ pub mod stdintf {
 pub struct LocalConnection {
     channel: Channel,
     filters: RefCell<Filters<LocalFilterCb>>,
+    all_signal_matches: bool,
 }
 
 /// A connection to D-Bus, non-async version where callbacks are Send but not Sync.
 pub struct Connection {
     channel: Channel,
     filters: RefCell<Filters<FilterCb>>,
+    all_signal_matches: bool,
 }
 
 /// A connection to D-Bus, Send + Sync + non-async version
 pub struct SyncConnection {
     channel: Channel,
-    filters: Mutex<Filters<SyncFilterCb>>
+    filters: Mutex<Filters<SyncFilterCb>>,
+    all_signal_matches: bool,
 }
 
 use crate::blocking::stdintf::org_freedesktop_dbus;
@@ -173,6 +176,10 @@ impl $c {
 
     /// Adds a new match to the connection, and sets up a callback when this message arrives.
     ///
+    /// If multiple [`MatchRule`]s match the same message, then by default only one of them will get
+    /// the callback. No guarantee is made about which one. This behaviour can be changed for signal
+    /// messages by calling [`all_signal_matches`](Self::all_signal_matches).
+    ///
     /// The returned value can be used to remove the match. The match is also removed if the callback
     /// returns "false".
     pub fn add_match<S: ReadAll, F>(&self, match_rule: MatchRule<'static>, f: F) -> Result<Token, Error>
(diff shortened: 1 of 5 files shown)


dbus/src/blocking.rs
```diff
@@ -204,20 +211,46 @@ impl $c {
         self.remove_match_no_cb(&mr.match_str())
     }
 
+    /// If true, configures the connection to send signal messages to all matching [`MatchRule`]
+    /// filters added with [`add_match`](Self::add_match) rather than just the first one. This will
+    /// result in the messages being duplicated, and so the message serial will be lost, but this is
+    /// generally not a problem for signals.
+    ///
+    /// This is false by default, for a newly-created connection.
+    pub fn all_signal_matches(&mut self, all_signal_matches: bool) {
```
diwic, line 220: Should be named `set_all_signal_matches`

dbus/src/blocking.rs
```diff
@@ -204,20 +211,46 @@ impl $c {
...
         if let Some(msg) = self.channel.blocking_pop_message(timeout)? {
-            let ff = self.filters_mut().remove_matching(&msg);
-            if let Some(mut ff) = ff {
-                if ff.2(msg, self) {
-                    self.filters_mut().insert(ff);
+            if self.all_signal_matches && msg.msg_type() == MessageType::Signal {
+                // If it's a signal and the mode is enabled, send a copy of the message to all
+                // matching filters.
+                for mut ff in self.filters_mut().remove_all_matching(&msg) {
+                    if let Ok(copy) = msg.duplicate() {
```
diwic, line 235: That all matching signals are removed during the callback process is a bit...interesting. It means that 1) if one callback panics, the remaining ones will be dropped and 2) one callback can remove earlier matches in the chain but not later ones, which is potentially confusing. 

That said; as long as this is properly documented as gotchas with the `all_signal_matches` mode, I guess that could be okay for the time being - if it confuses someone else they can write a patch to improve the situation.

dbus/src/message.rs
```diff
@@ -59,6 +59,20 @@ impl Message {
         Message { msg: ptr}
     }
 
+    /// Creates a new message that is a replica of this message, but without a serial.
+    ///
+    /// May fail if out of memory or file descriptors.
+    pub fn duplicate(&self) -> Result<Self, String> {
+        let ptr = unsafe {
```
diwic, line 66: Is there a reason this function is not called `copy`?

dbus/src/nonblock.rs
```diff
@@ -270,6 +294,16 @@ impl $c {
...
     }
+
+    /// If true, configures the connection to send signal messages to all matching [`MatchRule`]
+    /// filters added with [`add_match`](Self::add_match) rather than just the first one. This will
+    /// result in the messages being duplicated, and so the message serial will be lost, but this is
+    /// generally not a problem for signals.
+    ///
+    /// This is false by default, for a newly-created connection.
+    pub fn all_signal_matches(&mut self, all_signal_matches: bool) {
+        self.all_signal_matches = all_signal_matches;
```
diwic, line 305: Ditto, `set_all_signal_matches`

dbus/src/blocking.rs
diwic: Perhaps something like "when the all_signal_match mode is enabled, removing other matches inside a match callback is not supported".

diwic (commented):

In general I'm happy with the idea as well as the implementation, but see the review comments for some details.

Btw, I tried to think of some better name than `all_signal_matches` but I couldn't think of any.

dbus/src/message.rs
qwandor: Calling it `copy` would suggest to me it is something related to `std::marker::Copy`, which it is definitely not.

dbus/src/blocking.rs
qwandor: Added to the docstring for `set_all_signal_matches`, is there anywhere else you think I should mention it?

dbus/src/blocking.rs
qwandor: Done.

dbus/src/nonblock.rs
qwandor: Done.

qwandor:
Yeah, I'm not entirely happy with the name either, but couldn't think of anything better that wasn't ridiculously long.

diwic:
@qwandor  Okay, so I merged this as it is and added https://github.com/diwic/dbus-rs/commit/ebc07100d5d5934021849b5c7657388d76d524ec on top. Let me know if this commit looks good to you, and if you need a release to crates.io. Thanks!

qwandor:
Looks good. And yes, a release would be helpful thanks.

diwic:
@qwandor Released now. Enjoy!

PR #352, opened 2021-09-26
Allow signal_match_mode to be set for dbus-tokio connections

### LINKED ISSUE #346: If multiple match rules match a message, it only gets sent to the first one

qwandor:
I missed this in #351 unfortunately. Actually fixes #346 this time.

### DIFF
--- a/dbus/src/blocking.rs
+++ b/dbus/src/blocking.rs
@@ -7,6 +7,7 @@ use crate::{channel, Error, Message};
 use crate::message::{MatchRule, SignalArgs, MessageType};
 use crate::channel::{Channel, BusType, Token};
 use std::{cell::RefCell, time::Duration, sync::Mutex};
+use std::sync::atomic::{AtomicBool, Ordering};
 use crate::filters::Filters;
 
 #[allow(missing_docs)]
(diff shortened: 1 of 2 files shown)


qwandor:
This is a breaking API change for `dbus-tokio` because it adds a new parameter to existing functions. If you'd rather I added new functions for each variant I can do that again, but it ends up being a lot of different variations on almost the same thing. Unfortunately it's also a breaking change for `dbus` because I had to move the `set_signal_match_mode` method to a trait.

diwic:
What if we instead made `pub fn set_signal_match_mode` take a `&self` instead of a `&mut self`? That would not be a breaking change, and people could call `set_signal_match_mode` on the `Arc<LocalConnection>` or `Arc<SyncConnection>`, so you would not need to add anything to the trait either. 

This would mean changing the `all_signal_matches` variable from a `bool` to a `Cell<bool>` or `AtomicBool`, but that is not much of a problem either.

qwandor:
That might work. I'll have a try tomorrow.

qwandor:
Done, PTAL. Thanks for the suggestion, I think this is a better approach.

diwic:
Yeah, this is much less invasive. Looks good except for the ordering, either do acquire for load and release for store, or relaxed for both.

qwandor:
Done.

diwic:
@qwandor Okay, but have you tested it enough now to know that there aren't more things that need to go into that release to solve your problem?

qwandor:
I thought I had, but it turns out I introduced a bug when I was cleaning up. Sending another PR.

PR #354, opened 2021-09-29
Add convenience methods for building MatchRules

not-jan:
I've added a `MatchRuleBuilder` to employ the builder pattern for building MatchRules possibly avoiding a mut in the code for people using this library. I've also added a macro which is gated behind the `macros` feature since it adds a new dependency. It works like this: 

```rust
let mr: MatchRule<'static> = match_rule!(
    interface = "org.freedesktop.DBus.Properties",
    member = "PropertiesChanged",
    path = "/org/mpris/MediaPlayer2"
);
```

### DIFF
--- a/dbus/src/message/matchrule.rs
+++ b/dbus/src/message/matchrule.rs
@@ -116,4 +116,55 @@ impl<'a> MatchRule<'a> {
             _more_fields_may_come: (),
         }
     }
+
+    /// Enables eavesdropping for the generated message.
+    /// You probably want to use [BecomeMonitor](https://dbus.freedesktop.org/doc/dbus-specification.html#bus-messages-become-monitor) instead
+    pub fn with_eavesdrop(mut self) -> Self {
+        self.eavesdrop = true;
+        self
+    }
+
+    /// Sets the MatchRule to match on the message sender
+    pub fn with_sender(mut self, sender: impl Into<BusName<'a>>) -> Self {
+        self.sender = Some(sender.into());
+        self
+    }
+
+    /// Sets the MatchRule to match on the message sender and be strict
+    pub fn with_strict_sender(mut self, sender: impl Into<BusName<'a>>) -> Self {
+        self.sender = Some(sender.into());
+        self.strict_sender = true;
+        self
+    }
+
+    /// Sets the MatchRule to match on the message path and treat it as a namespace
+    pub fn with_namespaced_path(mut self, path: impl Into<Path<'a>>) -> Self {
+        self.path = Some(path.into());
+        self.path_is_namespace = true;
+        self
+    }
+
+    /// Sets the MatchRule to match on the message path
+    pub fn with_path(mut self, path: impl Into<Path<'a>>) -> Self {
+        self.path = Some(path.into());
+        self
+    }
+
+    /// Sets the MatchRule to match on the message interface
+    pub fn with_interface(mut self, intf: impl Into<Interface<'a>>) -> Self {
+        self.interface = Some(intf.into());
+        self
+    }
+
+    /// Sets the MatchRule to match on the message member
+    pub fn with_member(mut self, member: impl Into<Member<'a>>) -> Self {
+        self.member = Some(member.into());
+        self
+    }
+
+    /// Sets the MatchRule to match on the message type. This will usually be `"signal"`
+    pub fn with_type(mut self, ty: MessageType) -> Self {
+        self.msg_type = Some(ty);
+        self
+    }
 }


diwic:
Hi and thanks for the PR!

I can see the use for `.with_xxx()` builder-style methods, but they can just be implemented on `MatchRule` itself, there is no need for a builder struct. Thus your example would become:

    let mr = MatchRule::new_signal("org.freedesktop.DBus.Properties", "PropertiesChanged")
        .with_path("/org/mpris/MediaPlayer2");

...which is shorter (and does not miss the fact that you likely only want to match signal messages).

diwic:
As for the macro. I wonder if it's such an ergonomic improvement that it's worth making one, but if it is, I think you can make one without the dependency on `paste`? Also it should mandate the message type. So that your example desugars to something like:

    {
        let mut mr = MatchRule::new();
        mr.message_type = MessageType::Signal;
        mr.path = "/org/mpris/MediaPlayer2".into();
        mr.interface = "org.freedesktop.DBus.Properties".into();
        mr.member = "PropertiesChanged".into();
        mr
    }

not-jan:
> As for the macro. I wonder if it's such an ergonomic improvement that it's worth making one, but if it is, I think you can make one without the dependency on `paste`? Also it should mandate the message type. So that your example desugars to something like:
> 
> ```
> {
>     let mut mr = MatchRule::new();
>     mr.message_type = Some(MessageType::Signal);
>     mr.path = Some("/org/mpris/MediaPlayer2".into());
>     mr.interface = Some("org.freedesktop.DBus.Properties".into());
>     mr.member = Some("PropertiesChanged".into());
>     mr
> }
> ```

I agree with your concerns over the paste crate and frankly share them. I've discussed this with a few people on the Rust discord. One alternative would've been to use `std::concat_idents!` instead of the paste crate. This has the downside though that it is nightly only (and probably will stay that way indefinitely) and it can only deal with existing identifiers so the macro would have to be rewritten from the ground up. 

An alternative I was considering was adding a full fletched MatchRule parser that would allow us to use MatchRule literals straight from experimenting with e.g. dbus-monitor. This would however be a much larger code change and data flow wise it'd be kinda weird to parse a string into the structure just to unparse it and feed it to dbus again. Maybe the better choice here would be some kind of `const` MatchRule that can be build using a (proc-)macro?

> I can see the use for `.with_xxx()` builder-style methods, but they can just be implemented on `MatchRule` itself, there is no need for a builder struct. Thus your example would become:
> 
> ```
> let mr = MatchRule::new_signal("org.freedesktop.DBus.Properties", "PropertiesChanged")
>     .with_path("/org/mpris/MediaPlayer2");
> ```
> 
> ...which is shorter (and does not miss the fact that you likely only want to match signal messages).

I like that! The separate struct mainly stems from it originating in my crate :) 

I'll try implementing this one and the other suggestion later today.

diwic:
> One alternative would've been to use std::concat_idents! instead of the paste crate. 

You don't need to concat any idents if you do the assignment directly (instead of using `with_` like I suggested. Or did I miss anything?

 > it'd be kinda weird to parse a string into the structure just to unparse it and feed it to dbus again.

Well, it's not just for unparsing it, dbus-rs uses the matchrule struct for actual matching as well :-)

 > Maybe the better choice here would be some kind of const MatchRule that can be build using a (proc-)macro?

Not sure what this means, but if you want to make some functions `const` in order for this to work, go ahead - I see no disadvantages in doing so.

 > I'll try implementing this one and the other suggestion later today.
 
How about just implementing the `with_` functions as a first PR? I can merge that without much discussion or controversy.

The macro might need more discussion rounds before merging.

not-jan:
> How about just implementing the `with_` functions as a first PR? I can merge that without much discussion or controversy.

Done :)



> Not sure what this means, but if you want to make some functions `const` in order for this to work, go ahead - I see no disadvantages in doing so.

I've just checked and making the fn's const isn't quite as easy as expected. To elaborate on my idea: I was thinking about adding a match rule macro that works like this: ```rust
let mr: MatchRule<'static> = match_rule!("interface='org.freedesktop.Notifications',member='Notify'");
```

diwic:
> > How about just implementing the `with_` functions as a first PR? I can merge that without much discussion or controversy.
> 
> Done :)

@not-jan
Thanks, the code looks good, but could you squash it into one commit before I merge it? Thanks!

diwic:
> > 
> 
> I've just checked and making the fn's const isn't quite as easy as expected. To elaborate on my idea: I was thinking about adding a match rule macro that works like this:
> 
> ```rust
> let mr: MatchRule<'static> = match_rule!("interface='org.freedesktop.Notifications',member='Notify'");
> ```

This sounds like something that could be done through a regular function rather than a macro, and I'll be happy to review and merge such a function. I understand that maybe evaluating it compile time could be nice from a performance and correctness perspective, but I don't think parsing it runtime would make a noticeable difference in performance.

not-jan:
> > > How about just implementing the `with_` functions as a first PR? I can merge that without much discussion or controversy.
> > 
> > 
> > Done :)
> 
> @not-jan Thanks, the code looks good, but could you squash it into one commit before I merge it? Thanks!

Done :)

> This sounds like something that could be done through a regular function rather than a macro, and I'll be happy to review and merge such a function. I understand that maybe evaluating it compile time could be nice from a performance and correctness perspective, but I don't think parsing it runtime would make a noticeable difference in performance.

I'll try to come up with a design that works in a const context. Maybe even just to challenge myself :P

PR #356, opened 2021-10-03
Add MatchRule parsing 

not-jan:
As mentioned in #354 I've added MatchRule parsing. I'm not quite done yet and will probably revisit and refactor some parts but wanted some feedback first. 

Usage: 
```rust
let mr = MatchRule::parse(r"interface='org.freedesktop.Notifications',member='Notify'")?;
```

### DIFF
--- a/dbus/src/message/matchrule.rs
+++ b/dbus/src/message/matchrule.rs
@@ -1,5 +1,6 @@
 use crate::{Message, MessageType};
 use crate::strings::{BusName, Path, Interface, Member};
+use crate::message::parser;
 
 #[derive(Clone, Debug, Default)]
 /// A "match rule", that can match Messages on its headers.
@@ -60,7 +61,7 @@ impl<'a> MatchRule<'a> {
             if let Some(ref p) = msg.path() {
                 if x != p {
                     if self.path_is_namespace {
-                        p.starts_with(&**x) && &p[x.len()..x.len()+1] == "/"
+                        p.starts_with(&**x) && &p[x.len()..x.len() + 1] == "/"
                     } else { false }
                 } else { true }
             } else { false }
@@ -69,17 +70,17 @@ impl<'a> MatchRule<'a> {
 
     /// Returns whether or not the message matches the rule.
     pub fn matches(&self, msg: &Message) -> bool {
-        if let Some(x) = self.msg_type { if x != msg.msg_type() { return false; }};
+        if let Some(x) = self.msg_type { if x != msg.msg_type() { return false; } };
 
         if let Some(ref x) = self.sender {
             if let Some(s) = msg.sender() {
                 let check = self.strict_sender || (s.starts_with(":") == x.starts_with(":"));
-                if check && s != *x { return false }
-            } else if self.strict_sender { return false }
+                if check && s != *x { return false; }
+            } else if self.strict_sender { return false; }
         };
-        if !self.path_match(msg) { return false }
-        if self.interface.is_some() && msg.interface() != self.interface { return false };
-        if self.member.is_some() && msg.member() != self.member { return false };
+        if !self.path_match(msg) { return false; }
+        if self.interface.is_some() && msg.interface() != self.interface { return false; };
+        if self.member.is_some() && msg.member() != self.member { return false; };
         true
     }
 
@@ -166,5 +167,11 @@ impl<'a> MatchRule<'a> {
     pub fn with_type(mut self, ty: MessageType) -> Self {
         self.msg_type = Some(ty);
         self
+    }
+
+    /// Tries parsing a MatchRule from a String. Please note however that not all features supported
+    /// by DBus are supported by dbus-rs (yet). args and destinations are not supported yet.
+    pub fn parse(text: &'a str) -> Result<Self, parser::Error> {
+        parser::Parser::new(text)?.parse()
     }
 }
--- a/dbus/src/message/parser.rs
+++ b/dbus/src/message/parser.rs
@@ -1,0 +1,261 @@
+use std::error::Error as stdError;
+use std::fmt::{Debug, Formatter, Display};
+use crate::message::MatchRule;
+use crate::{MessageType, Path};
+use std::convert::TryFrom;
+use crate::strings::{Interface, BusName, Member};
+
+// Our grammar:
+// rules: rule (, rule)*
+// rule: sender | type | interface | member | path | path_namespace | destination | arg | arg_path
+// bool : 'true' | 'false'
+// type: "type" "=" message_type
+// message_type: "'signal'" | "'method_call'" | "'method_return'" | "'error'"
+// sender: "sender" "=" string
+// interface: "interface" "=" string
+// member: "member" "=" string
+// path: "path" "=" string
+// path_namespace: "path_namespace" "=" string
+// destination: "destination" "=" string
+// arg: "arg" 0-63 "=" string
+// arg_path: "arg" 0-63 "path" "=" string
+// eavesdrop: "eavesdrop" "=" bool
+
+
+#[derive(Clone, Debug)]
+/// Error type that covers errors that might happen during parsing.
+pub enum Error {
+    /// The type specified in the match rule is unknown
+    UnknownType,
+    /// The key is wrong / unsupported
+    UnknownKey,
+    /// Boolean could not be parsed
+    BadBoolean,
+    /// Error that occured while converting a string to a DBus format
+    BadConversion(String),
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Error while parsing MatchRule: ")?;
+        match self {
+            Error::UnknownType => {
+                write!(f, "Unsupported message type")
+            }
+            Error::UnknownKey => {
+                write!(f, "Unknown key used")
+            }
+            Error::BadBoolean => {
+                write!(f, "Got bad boolean value")
+            }
+            Error::BadConversion(err) => {
+                write!(f, "Error while converting: {}", err)
+            }
+        }
+    }
+}
+
+impl stdError for Error {}
+
+/// Key-Value-pair
+pub type TokenRule<'a> = (&'a str, &'a str);
+/// Fixed size buffer for match rule tokens
+pub type TokenBuffer<'a> = Vec<TokenRule<'a>>;
+
+#[derive(Clone, Debug)]
+/// Tokenizes a match rule into key-value-pairs
+struct Tokenizer<'a> {
+    text: &'a str,
+}
+
+impl<'a> Tokenizer<'a> {
+    /// Builds a new tokenizer for the input &str
+    pub fn new(text: &'a str) -> Self {
+        Self {
+            text,
+        }
+    }
+
+    /// Parse the key part of the key-value-pair. This is rather easy as all keys are rather
+    /// easily defined and may only contain `[a-z0-9]` so we can simply split at `'='`.
+    fn key(&self) -> (&'a str, &'a str) {
+        let index = self.text.find('=').unwrap_or_else(|| self.text.len());
+
+        ((&self.text[..index]).trim(), &self.text[index + 1..])
+    }
+
+    /// Parses values as generic strings.
+    /// This does not do any validation (yet) with regards to supported characters.
+    fn value(&self) -> (&'a str, &'a str) {
+        let mut i = 0;
+        let mut quoted = false;
+        let mut escape = false;
+
+        for c in self.text.chars() {
+            match c {
+                '\'' if !escape => {
+                    quoted = !quoted;
+                }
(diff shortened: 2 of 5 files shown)


dbus/Cargo.toml
```diff
@@ -19,6 +19,7 @@ libdbus-sys = { path = "../libdbus-sys", version = "0.2.2" }
 futures-util = { version = "0.3", optional = true, default-features = false }
 futures-channel = { version = "0.3", optional = true }
 futures-executor = { version = "0.3", optional = true }
+arrayvec = { version = "0.7", optional = true }
 # dbus-native-channel = { path = "../dbus-native-channel", version = "0.1", optional = true }
```
diwic, line 23: Could you give a motivation to your use of `arrayvec` (instead of using an ordinary `Vec`)? If it's just optimization, then I would argue it's premature optimization.

diwic: Using an ordinary vec would not add any additional dependencies, meaning that we don't have to think about features. It just seems simpler to me.

dbus/src/message/matchrule.rs
```diff
@@ -167,4 +167,300 @@ impl<'a> MatchRule<'a> {
...
+    use std::error::Error as stdError;
+    use arrayvec::ArrayVec;
+    use std::fmt::{Debug, Formatter, Display};
+    use crate::message::MatchRule;
+    use crate::MessageType;
+    use std::convert::TryFrom;
+    use crate::arg::RefArg;
+
+    const MAX_RULE_TOKENS: usize = 16;
+
```
diwic, line 207: There is no word about maximum rule tokens in the D-Bus specification as I can see...? Where does this restriction come from?

dbus/src/message/matchrule.rs
```diff
@@ -167,4 +167,300 @@ impl<'a> MatchRule<'a> {
...
+    /// Tries parsing a MatchRule from a String. Please note however that not all features supported
+    /// by DBus are supported by dbus-rs (yet). args and destinations are not supported yet.
+    pub fn parse(text: &'a str) -> Result<Self, parser::Error> {
+        parser::Parser::new(text)?.parse()
+    }
+}
+
+
+#[cfg(feature = "parser")]
+pub mod parser {
```
diwic, line 181: How about moving this module to its own file?

dbus/src/message/matchrule.rs
```diff
@@ -167,4 +167,300 @@ impl<'a> MatchRule<'a> {
...
+        }
+
+        /// Cleans a string from the string syntax allowed by DBus. This includes concatenating
+        /// things like `''\'''` to `'`. DBus strings sort of work like in a POSIX shell and
+        /// concatenation is implied. There is only one escape sequence and that is in an unquoted
+        /// substring a backslash may escape an ASCII apostrophe (U+0027).
+        /// This method is the only one within the parser that allocates and in theory could be
+        /// rewritten to taking a `&mut str` instead of returning a `String` since all
+        /// strings are |output| <= |buf.len()|.
+        fn clean_string(&self, buf: &str) -> String {
```
diwic, line 357: I guess the easiest optimization here would be to have a clean_string check (this could be part of the `value` function if you like) that indicates whether or not the string actually needs to be cleaned; the common case would be that it does not need cleaning and so one could skip calling this function in the first place.

dbus/src/message/matchrule.rs
```diff
@@ -167,4 +167,300 @@ impl<'a> MatchRule<'a> {
...
+            for &(key, raw_value) in &self.tokens {
+                let value = self.clean_string(raw_value);
+                match key {
+                    "type" => {
+                        match_rule = match_rule.with_type(MessageType::try_from(value.as_str())?);
+                        Ok(())
+                    }
+                    "interface" => {
+                        match_rule = match_rule.with_interface(value);
+                        Ok(())
```
diwic, line 396: The `with_xxx` functions are quick and ergonomic; but panics if the path/interface/etc contains invalid characters. A parser could do better by constructing the path/interface first using a fallible function and return an error in case the path/interface/etc is invalid.

diwic (commented):

In addition to the review comments (which can be rather dry), I'd like to add that in general I like what I'm seeing, so see the comments as possible improvements rather than critique, and keep up the good work :-)

dbus/src/message/matchrule.rs
not-jan: I took it from the DBus source code: https://github.com/freedesktop/dbus/blob/ef55a3db0d8f17848f8a579092fb05900cc076f5/bus/signals.c#L720

dbus/Cargo.toml
not-jan: This was mostly done in preparation to *eventually* move things to a `const` context. As mentioned below there are a finite amount of rules that can actually be matched upon (and even less so in the current implementation) so this is an easy allocation to not do.

dbus/src/message/matchrule.rs
not-jan: To be honest when I started writing the parser I was unsure how much work it would actually entail. I'll move it in the next commit.

dbus/src/message/matchrule.rs
not-jan: I've mentioned this in the tests but ideally I'd like to implement TryInto here using the `dbus_validate_` family of functions.

dbus/src/message/matchrule.rs
not-jan: Good point, I'll get on that. I've designed my parser with the worst case scenarios in mind e.g. `arg0=''\''',arg1='\',arg2=',',arg3='\\'` so I've written it in a rather defensive code style. I'm not all too happy with this method in general and the comment isn't quite truthful either since we'll usually be working with &'static str and not &'static mut str

dbus/src/message/matchrule.rs
diwic: Hmm, and above that was a FIXME saying:

 >  * FIXME this is sort of busted now with arg matching, but we let
 > * you match on up to 10 args for now

So they're basically saying that the limit of 16 should be lifted. Which means we should probably not have such an upper limit.

dbus/Cargo.toml
diwic: Moving into that eventual const context is going to be difficult, if not impossible. The D-Bus library works with zero-terminated strings, and so that is the representation of paths, interfaces, etc in dbus-rs as well (even though if I ever finish off that native implementation it would make sense to have `&str` instead of `&CStr`). 

Which means that copying the strings once is almost inevitable, because you need to add that `\0` afterwards somehow.

dbus/src/message/matchrule.rs
diwic: If you call `Path::new` or `Path::from_slice`, it will call into `dbus_validate_path` for you already. (And similar for `Interface` etc.)

There is no TryFrom/TryInto because nobody has implemented that and I wrote that code before TryFrom existed in std. Feel free to do so if you like.

dbus/src/message/matchrule.rs
not-jan: For us to get anywhere near the 16 we'd have to implement arg matching first :P But I see where you're coming from. Perhaps we should figure out a way to implement arg matching first.

dbus/src/message/matchrule.rs
not-jan: I just tried implementing it and failed due to the blanket implementation in core :( Basically it implements TryInto<U> for all types that implement Into<U> to help with interfaces. And From implicitly implements Into so we're out of luck unless we switch from From to TryFrom which would break a lot of usages but would eliminate the panic!'s.

Relevant issue: https://github.com/rust-lang/rust/issues/50133

dbus/src/message/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+// eavesdrop: "eavesdrop" "=" bool
+
+/// The maximum rule elements supported by DBus
+/// This is taken from https://github.com/freedesktop/dbus/blob/master/bus/signals.c#L720
+const MAX_RULE_TOKENS: usize = 16;
+
+#[derive(Clone, Debug)]
+/// Error type that covers errors that might happen during parsing.
+pub enum Error {
+    /// DBus has a limit on the size of match rules.
```
diwic, line 32: This needs to be publicly exported somehow, e g as `dbus::message::MatchRuleParserError` or something shorter.

dbus/src/message/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+                write!(f, "Error while converting: {}", err)
+            }
+        }
+    }
+}
+
+impl stdError for Error {}
+
+impl<'a> TryFrom<&'a str> for MessageType {
+    type Error = Error;
```
diwic, line 71: It is a bit incoherent that a conversion from `&str` to `MessageType` would give a potential `MatchRuleParserError`.
Either move this to `message.rs` with `type Error = ()` and do a `map_err` in `parser.rs`, or skip the tryfrom and use an ordinary method instead.

dbus/src/message/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+// path_namespace: "path_namespace" "=" string
+// destination: "destination" "=" string
+// arg: "arg" 0-63 "=" string
+// arg_path: "arg" 0-63 "path" "=" string
+// eavesdrop: "eavesdrop" "=" bool
+
+/// The maximum rule elements supported by DBus
+/// This is taken from https://github.com/freedesktop/dbus/blob/master/bus/signals.c#L720
+const MAX_RULE_TOKENS: usize = 16;
+
```
diwic, line 28: My objection about features, arrayvec and MAX_RULE_TOKENS remains. It is simpler just to use an ordinary vec and not impose unneeded limits.

dbus/src/message/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+            i += 1;
+        }
+
+        // Skip comma if there is still space in the buffer
+        if self.text.len() == i {
+            ((&self.text[..i]).trim(), &self.text[i..])
+        } else {
+            ((&self.text[..i]).trim(), &self.text[i + 1..])
+        }
```
diwic, line 143: Less copy-paste with

    // Skip comma if there is still space in the buffer
    let j = if self.text.len() == i { i } else { i + 1 };
    (&self.text[..i]).trim(), &self.text[j..])

dbus/src/message/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+        assert_eq!(mr.match_str(), "interface='org.freedesktop.Notifications',member='Notify'");
+        let mr = MatchRule::parse(r"interface='org.mpris.MediaPlayer2.Player' , path= /org/mpris/MediaPlayer2,member='Notify', eavesdrop ='true'")?;
+        assert_eq!(mr.match_str(), "path='/org/mpris/MediaPlayer2',interface='org.mpris.MediaPlayer2.Player',member='Notify',eavesdrop='true'");
+        Ok(())
+    }
+
+    #[test]
+    fn test_malformed() {
+        // This panics due to `cstring_wrapper!` not supporting TryInto / TryFrom
+        assert!(MatchRule::parse(r"interface='org.freedesktop.Notifications',member=").is_err());
```
diwic, line 283: I don't understand this comment. Is it still relevant?

dbus/src/message/matchrule.rs
diwic: Oh, ok. Yeah switching from `From` to `TryFrom` is probably too backwards incompatible at this point...maybe for 0.10 if we ever release that.

dbus/src/message/parser.rs
not-jan: Oops, forgot to remove that comment.

dbus/src/message/parser.rs
not-jan: Considering that a `const` future is far, far away I'll agree.

dbus/src/message/parser.rs
not-jan: Agree, I've moved it to message.rs and changed the error to ().

dbus/src/message/parser.rs
not-jan: I think I've done this but I'm not 100% sure if it works like I want it to. Can you maybe double check?

diwic: It looks as I expect it to. A tip if you want to double check is to run `cargo doc --open --no-deps` and check the `message` module to see if it shows up there.

diwic:
@not-jan After your latest changes I think the code looks good enough for merging (after being squashed into one commit).

Are you planning additional changes or are you satisfied with the code the way it is now?

diwic:
(except the gitignore change which can be either removed or stay in its own commit)

not-jan:
I think it's good enough. I don't think I can get rid of the merge commit though.

diwic:
Applied manually in b7c22934334 after having removed the .gitignore change.

PR #357, opened 2021-10-05
Dbus: Allow 25 function arguments instead of 13

GreenArchon:
This limit was increased in 389d41066e2dddfd62719282519121ff45b9f3d2,
however this is still insufficient for some methods (ie. org.gpsd's fix
signal).

Obviously this is just pushing an arbitrary limit further back, but I
figured running out of letters in the alphabet was a good place to stop
for now. Bonus points for finding the letter that couldn't be used.


diwic:
🙈

That interface of gpsd, they should really refactor it into a dictionary...

Oh well. I guess I have to merge this. I hope compilation time or code size does not significantly increase because of this?

GreenArchon:
> Oh well. I guess I have to merge this.

Thanks :grimacing: .

> I hope compilation time or code size does not significantly increase because of this?

From quick tests in release mode, it does affect the size of libdbus.rlib a little bit (6.5MB instead of 6.0MB), but not of final executable linking to it, which was identical give or take a few bytes. So dead code isn't linked.

As for compile time, it seems to be ~0.2s slower on my machine.

diwic:
@GreenArchon Okay. Thanks for checking!

Issue #363, opened 2021-12-03
Help with building a struct from a value in `PropMap` (crossroads)


Toqozz:
Hey, I searched around and tried to find some examples relating to this but couldn't find anything, apologies if there's some information on how to do this somewhere.

I'm trying to implement the [Desktop Notifications Specification](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html), specifically I'm trying to grab the `image-data` hint from the `hints` component.  The signature for this property is `iiibiiay`.

This is one of my attempts:
```rust
pub struct DBusImage {
    pub width: i32,
    pub height: i32,
    pub rowstride: i32,
    pub one_point_two_bit_alpha: bool,
    pub bits_per_sample: i32,
    pub channels: i32,
    pub data: Vec<u8>,
}

impl dbus_codegen::OrgFreedesktopNotifications for Notify {
    ...

    fn notify(&mut self, app_name: String, replaces_id: u32, app_icon: String, summary: String, body: String, actions: Vec<String>, hints: arg::PropMap, expire_timeout: i32) -> Result<u32, dbus::MethodErr> {
        let var = hints.get("image-data");
        if let Some(img) = var {
            if let Some(mut iter) = img.0.as_iter() {
                // Just trying to cast one thing as an example, but maybe there is a way to easily pull everything into the struct?
                // error: cast requires that `hints` is borrowed for `'static`
                let w = arg::cast::<i32>(iter.next().unwrap());
            }
        }

        Ok(0)
    }
    ...
}
```

I was looking at the [argument guide](https://github.com/diwic/dbus-rs/blob/master/dbus/examples/argument_guide.md), but most of the examples are for doing `value.read()` into the specific variant.  I noticed that I can probably do `value.as_i64().unwrap() as i32` or something like that, but I feel like I'm just completely missing something.

I should mention also that I had this working before with  `dbus-tree` and am trying to migrate over.  Here's what I did previously:
```rust
impl<'a> arg::Get<'a> for DBusImage {
    fn get(i: &mut arg::Iter) -> Option<Self> {
        let width = i.read::<i32>().ok()?;
        let height = i.read::<i32>().ok()?;
        let rowstride = i.read::<i32>().ok()?;
        let one_point_two_bit_alpha = i.read::<bool>().ok()?;
        let bits_per_sample = i.read::<i32>().ok()?;
        let channels = i.read::<i32>().ok()?;
        let data = i.read::<Vec<u8>>().ok()?;

        Some(
[...]

PR #365, opened 2021-12-07
Struct/PropMap example

Toqozz:
From discussion in #363.

Added an example for Struct casting.  Also includes some example code/explanation regarding `PropMap`.

I'm not sure if it's in the right place or not, but it seems to fit.  Sorry for the length, but I wanted to be cover a few issues that I ran into along the way.


dbus/examples/argument_guide.md
```diff
@@ -147,7 +147,69 @@ Structs
...
+    replaces_id: u32,
+    app_icon: String,
+    summary: String,
+    body: String,
+    actions: Vec<String>,
+    hints: arg::PropMap,
+    expire_timeout: i32
+) -> Result<u32, dbus::MethodErr> {
+    // PropMap is just an alias for a HashMap:
+    // pub type PropMap = HashMap<String, Variant<Box<dyn RefArg + 'static>>>;
```
diwic, line 167: This comment can be removed, it's already stated in the previous paragraph.

dbus/examples/argument_guide.md
```diff
@@ -147,7 +147,69 @@ Structs
...
+
+    use dbus::arg::prop_cast;
+
+    // Retrieving a property from a PropMap of type byte (`u8`).
+    let urgency: Option<&u8> = prop_cast(&hints, "urgency");
+    //let urgency: Option<u8> = prop_cast(&hints, "urgency").copied();
+    //let urgency = prop_cast::<u8>(&hints, "urgency");    // Option<&u8>
```
diwic, line 174: These two comments could also be removed, not sure what they are for?

diwic (commented):
Looks good except for some superfluous comments. Thanks for contributing!

dbus/examples/argument_guide.md
Toqozz: Just wanted to show different ways of achieving the same thing.

PR #372, opened 2022-01-02
Fix links to tokio crossroads example

smheidrich:
Link is broken (404) - replaced it with what is probably the new path of the file it was meant to refer to.


diwic:
Thanks!

PR #376, opened 2022-02-03
Port dbus-rs to rustix.

sunfishcode:
This ports dbus-rs from using the libc crate directly to using rustix, a syscall
wrapper crate with a focus on safe APIs. This factors out some unsafe blocks and
manual error handling.

This PR doesn't completely replace libc, as dbus-rs uses several libc types in
its public API, and this PR seeks to avoid changing the public API.

### DIFF
--- a/dbus-native-channel/Cargo.toml
+++ b/dbus-native-channel/Cargo.toml
@@ -8,3 +8,4 @@ edition = "2018"
 
 [dependencies]
 libc = "0.2.80"
+rustix = "0.33.0"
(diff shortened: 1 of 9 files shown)


diwic:
Hi and thanks for the PR,

From just looking at the code, my gut feeling is that the disadvantages of adding a dependency on `rustix` (which in turn has a few other dependencies), is not worth the few advantages in the code that it provides.

It's worth mentioning that the `rtkit` example is where the cleanup is heaviest so there it could make sense to have it, and that `OwnedFd` is to be deprecated, as one can use `std::fs::File` as an fd container instead, so there is little use in trying to improve that anyway.

sunfishcode:
I understand; thanks for taking a look!

PR #378, opened 2022-03-14
codegen: avoid useless deprecation attributes in impls

sirhcel:
When generating bindings for interfaces with deprecated items, `dbus-codegen-rust` emits deprecation attributes even for impls. The generated code will fail to build with an actual Rust toolchain default `#[deny(useless_deprecated)]. For example:
```
error: this `#[deprecated]` annotation has no effect
   --> src/connman/manager.rs:190:5
    |
190 |     #[deprecated(note = "true")]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove the unnecessary deprecation attribute
    |
    = note: `#[deny(useless_deprecated)]` on by default
```

This PR introduces distinguishing between `trait` and `impl` output context and generates deprecation attributes only for the former. The generated code will now compile with `#[deny(useless_deprecated)]`.

### DIFF
--- a/dbus-codegen/src/generate/write.rs
+++ b/dbus-codegen/src/generate/write.rs
@@ -2,6 +2,12 @@ use super::*;
 
 use super::types::*;
 
+#[derive(Debug, PartialEq, Eq)]
+enum GenContext {
+    Trait,
+    Impl,
+}
+
 fn make_result(success: &str, opts: &GenOpts) -> String {
     if opts.methodtype.is_some() {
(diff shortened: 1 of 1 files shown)


diwic:
Merged now. Thanks for your contribution! And let me know if you need a release.

sirhcel:
Thank you for getting to this PR so quick! I'm not in a hurry with a release (as I already generated the bindings I wanted manually and have no backlog right now).

PR #380, opened 2022-03-30
crossroads: Remove requirement of having properties for object manager

sonnysasaka:
The D-Bus specification does not require properties to exist for an
object path to be advertised in ObjectManager
(https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager).


diwic:
I can't help to wonder what I was thinking when I wrote that...if there was a reason. Can't think of any right now so I'll just merge this one and hope that it's the correct decision. :-)

Thanks for your contribution!

sonnysasaka:
Thanks for the review!

Issue #386, opened 2022-06-01
Cannot dynamically add and remove interfaces from an existing path


sonnysasaka:
Currently Crossroads has some limitations:
* Cannot add an interface dynamically to an existing path.
* Cannot add multiple interfaces to a path having different backing
  data.
* Cannot remove a single interface from a path.

This limitation is not inherent from D-Bus spec, because the spec allows
those. A concrete usage of this pattern is by BlueZ (Bluetooth stack)
where it adds/removes interfaces to the same path dynamically based on the
availability of Bluetooth profiles on a remote device represented by
that path.

diwic:
Hi,

As for dynamically adding and removing interfaces from a path, I'm happy to take a PR that implements such functionality.

As for having different backing data for different interfaces, this seems like a bad idea in general so I don't want to implement it in the API, but for people that want it that way, they can let the backing data for the object be a `HashMap<IfaceToken, Box<dyn Any + Send>>` (use `String` or `dbus::strings::Interface` instead of `IfaceToken` if that is simpler), and then do the lookup in the method handlers when necessary.

sonnysasaka:
Hi diwic,

Thanks for your feedback. I created another pull request based on your feedback, i.e. I am adding `add_interface` and `remove_interface` without changing any backing data and the caller can share the same backing data if they need to. Please take a look at #388, thanks!

PR #387, opened 2022-06-01
WIP: A D-Bus object is a pair of path and interface

sonnysasaka:
The change proposed here is to change the internal concept of an object:
In D-Bus, an object can be interpreted as a pair of path and interface.

This interpretation implies that:
* An interface can be added dynamically to an existing path, backed by a
  different "data", without having to reset other existing interfaces on
  that path.
* An interface can be removed dynamically from an existing path without
  having to remove all other interfaces from the path.
* The org.freedesktop.DBus.Properties interface is a pseudo-interface,
  because the real interface it operates on is encoded in the .Get,
  .Set, and .GetAll first message argument.
* The Introspectable and ObjectManager are special interfaces which do
  not need backing "data".

This change requires some changes to the public interface, highlighted
below:
* `Crossroads::insert` accepts a single interface instead of multiple
  interfaces. This is because each interface may be backed by a
  different data type. Adding multiple interfaces can still be done by
  calling `Crossroads::insert` multiple times.
  An alternatives to this:
  * Allow calling `Crossroads::insert` with multiple interfaces if the
    data types are the same. This implies that the data has to be
    `Clone` or that the data is also provided as an `IntoIterator`.
* `Crossroads::remove` accepts an interface alongside the path. This
  allows a removal of a single path+interface pair.
  An alternative to this:
  * Also provide `Crossroads::remove_all_interfaces` that behaves like
    the old remove, which removes all interfaces of an object. This
    cannot return the removed data because there are multiple data with
    different types.

This patch is work in progress and is still missing tests. Please do not merge
this patch directly but let me know whether the idea looks good.

Fixes diwic/dbus-rs#386

### DIFF
--- a/dbus-crossroads/src/crossroads.rs
+++ b/dbus-crossroads/src/crossroads.rs
@@ -1,11 +1,12 @@
 use std::pin::Pin;
 use std::sync::Arc;
 use dbus::channel::Sender;
+use std::collections::btree_map::Entry;
 use std::future::Future;
 use std::marker::PhantomData;
 use crate::{Context, MethodErr, IfaceBuilder, stdimpl};
 use crate::ifacedesc::Registry;
-use std::collections::{BTreeMap, HashSet};
+use std::collections::{BTreeMap, HashMap};
 use std::any::Any;
 use std::fmt;
 use crate::utils::Dbg;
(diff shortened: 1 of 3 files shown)


diwic:
> In D-Bus, an object can be interpreted as a pair of path and interface.

Hi and thanks for the suggestion,

However, this is not my understanding of a D-Bus object. A D-Bus object is represented by its path, not the pair of path and interface. There can be many interfaces to the same object. This is also how interfaces work in programming languages in general.

sonnysasaka:
Hi diwic,

Thanks for the feedback. I created another pull request based on the approach that you suggested (#388) so I will close this one.

PR #396, opened 2022-09-29
Add docker-compose instructions

ln-12:
I struggled to configure all the paths needed for docker without cross, so I want to share my result here.


diwic:
Thanks!

PR #397, opened 2022-10-31
Provide command to interact with example.

dvogel:
While trying to get a basic IPC method established, it took me a while to figure out how the various strings used in the example could be strung together to test it with `dbus-send`. I learned the hard way that the `--print-reply` option is required to get output from either the example's stdout or from `dbus-send`. Hopefully this change saves others a little time.


diwic:
LGTM so I merged it. Although you can also use the `tokio_client.rs` example. Thanks!

PR #398, opened 2022-11-17
README.md: made crossroads example compile

komar007:
closure passed to b.method returns tuple instead of String


diwic:
Thanks!

PR #406, opened 2022-12-23
crossroads: Support "" as interface name in org.freedesktop.DBus.Properties.GetAll

zevweiss:
The spec [0] allows passing an empty string as the interface name for all methods of the org.freedesktop.DBus.Properties interface.  This implementation doesn't cover the Get or Set methods, but does at least add support for it in GetAll.

[0] https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-properties

### DIFF
--- a/dbus-crossroads/src/stdimpl.rs
+++ b/dbus-crossroads/src/stdimpl.rs
@@ -227,7 +227,19 @@ fn get(mut ctx: Context, cr: &mut Crossroads, (interface_name, property_name): (
     propctx.call_prop(cr, false).map(|propctx| { propctx.context.unwrap() })
 }
 
+fn getall_all(ctx: Context, cr: &mut Crossroads) -> Option<Context> {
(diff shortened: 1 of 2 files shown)


diwic:
The specification is not crystal clear here IMO but I believe your interpretation is a reasonable one. Thanks for the contribution!

zevweiss:
Agreed, there's an unfortunately lack of information in the spec on exactly what behavior an empty-string interface argument should produce; I referenced the libsdbus implementation [here](https://github.com/systemd/systemd/blob/20ada514648d2c479a49689c6714c72af31ab8f6/src/libsystemd/sd-bus/bus-objects.c#L1362-L1363) and [here](https://github.com/systemd/systemd/blob/20ada514648d2c479a49689c6714c72af31ab8f6/src/libsystemd/sd-bus/bus-objects.c#L842-L843).

PR #408, opened 2022-12-24
Add Vendored Libdbus Support

landhb:
This PR adds vendoring behind a feature flag, much like many other `sys` crates in the ecosystem.

- Allows users to build without having `libdbus-dev-1` installed.  Since it will be compiled from source.
- Allows users to cross compile with `cross` without needing a custom docker image.
- Uses the `cc` crate directly, so neither `cmake`  or `meson` are required for downstream users.

A downstream user would simply enable the `vendored` feature:

```
dbus = {version = "0.9.7", features = ["vendored"]}
```

And build with cross:

```
cross build --target x86_64-unknown-linux-musl
```

The PR is currently passing the existing test suite on x86_64:

```
cargo test --features vendored
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libdbus-sys/vendor/dbus"]
+	path = libdbus-sys/vendor/dbus
+	url = https://github.com/freedesktop/dbus.git
--- a/libdbus-sys/build.rs
+++ b/libdbus-sys/build.rs
@@ -1,8 +1,17 @@
 extern crate pkg_config;
+use std::error::Error;
 
-fn main() {
+#[cfg(feature = "vendored")]
+mod build_vendored;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // Invalidate the built crate whenever these files change
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=build_vendored.rs");
+
     // See https://github.com/joshtriplett/metadeps/issues/9 for why we don't use
     // metadeps here, but instead keep this manually in sync with Cargo.toml.
+    #[cfg(not(feature = "vendored"))]
     if let Err(e) = pkg_config::Config::new().atleast_version("1.6").probe("dbus-1") {
         eprintln!("pkg_config failed: {}", e);
         eprintln!(
@@ -15,4 +24,8 @@ fn main() {
         );
         panic!();
     }
+
+    #[cfg(feature = "vendored")]
+    build_vendored::build_libdbus()?;
+    Ok(())
 }
--- a/libdbus-sys/build_vendored.rs
+++ b/libdbus-sys/build_vendored.rs
@@ -1,0 +1,391 @@
+//! Vendored Libdbus build.
+//! 
+//! Configures + Compiles + Statically Links a pinned libdbus release.
+use std::collections::HashMap;
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::process::Command;
+use std::path::Path;
+use std::error::Error;
+use std::string::ToString;
+use std::collections::hash_map::Iter;
+
+// Version of vendored libdbus
+const DBUS_VERSION: &str = "\"1.14.4\"";
+const DBUS_VERSION_MAJOR: u32 = 1;
+const DBUS_VERSION_MINOR: u32 = 14;
+const DBUS_VERSION_MICRO: u32 = 4;
+
+// Location of dbus-daemon socket.
+// Systemd Default: https://github.com/systemd/systemd/blob/main/src/libsystemd/sd-bus/bus-internal.h#L20
+const DBUS_SYSTEM_BUS_DEFAULT_ADDRESS: &str = "/run/dbus/system_bus_socket";
+const DBUS_SESSION_SOCKET_DIR: &str = "/tmp";
+const DBUS_MACHINE_UUID_FILE: &str = "/etc/machine-id";
+
+/// Source Files Required
+/// Reference: https://github.com/freedesktop/dbus/blob/master/dbus/meson.build#L31
+/// Meson variable name: dbus_lib_sources
+const SOURCE_FILES: &'static [&'static str] = &[
+    "./vendor/dbus/dbus/dbus-address.c",
+    "./vendor/dbus/dbus/dbus-auth.c",
+    "./vendor/dbus/dbus/dbus-bus.c",
+    "./vendor/dbus/dbus/dbus-connection.c",
+    "./vendor/dbus/dbus/dbus-credentials.c",
+    "./vendor/dbus/dbus/dbus-errors.c",
+    "./vendor/dbus/dbus/dbus-keyring.c",
+    "./vendor/dbus/dbus/dbus-marshal-byteswap.c",
+    "./vendor/dbus/dbus/dbus-marshal-header.c",
+    "./vendor/dbus/dbus/dbus-marshal-recursive.c",
+    "./vendor/dbus/dbus/dbus-marshal-validate.c",
+    "./vendor/dbus/dbus/dbus-message.c",
+    "./vendor/dbus/dbus/dbus-misc.c",
+    "./vendor/dbus/dbus/dbus-nonce.c",
+    "./vendor/dbus/dbus/dbus-object-tree.c",
+    "./vendor/dbus/dbus/dbus-pending-call.c",
+    "./vendor/dbus/dbus/dbus-resources.c",
+    "./vendor/dbus/dbus/dbus-server-debug-pipe.c",
+    "./vendor/dbus/dbus/dbus-server-socket.c",
+    "./vendor/dbus/dbus/dbus-server.c",
+    "./vendor/dbus/dbus/dbus-sha.c",
+    "./vendor/dbus/dbus/dbus-signature.c",
+    "./vendor/dbus/dbus/dbus-syntax.c",
+    "./vendor/dbus/dbus/dbus-threads.c",
+    "./vendor/dbus/dbus/dbus-timeout.c",
+    "./vendor/dbus/dbus/dbus-transport-socket.c",
+    "./vendor/dbus/dbus/dbus-transport.c",
+    "./vendor/dbus/dbus/dbus-watch.c",
+];
+
+/// Internal files
+/// Reference: https://github.com/freedesktop/dbus/blob/master/dbus/meson.build#L67
+/// Meson variable name: dbus_shared_sources
+const INTERNAL_FILES: &'static [&'static str] = &[
+    "./vendor/dbus/dbus/dbus-dataslot.c",
+    "./vendor/dbus/dbus/dbus-file.c",
+    "./vendor/dbus/dbus/dbus-hash.c",
+    "./vendor/dbus/dbus/dbus-internals.c",
+    "./vendor/dbus/dbus/dbus-list.c",
+    "./vendor/dbus/dbus/dbus-marshal-basic.c",
+    "./vendor/dbus/dbus/dbus-memory.c",
+    "./vendor/dbus/dbus/dbus-mempool.c",
+    "./vendor/dbus/dbus/dbus-pipe.c",
+    "./vendor/dbus/dbus/dbus-string.c",
+    "./vendor/dbus/dbus/dbus-sysdeps.c",
+];
+
+/// Unix platform files
+/// References:
+/// - https://github.com/freedesktop/dbus/blob/master/dbus/meson.build#L139
+/// - https://github.com/freedesktop/dbus/blob/master/dbus/meson.build#L144
+/// Meson variable names: dbus_lib_sources + dbus_shared_sources
+const UNIX_SOURCES: &'static [&'static str] = &[
+    // Lib
+    "./vendor/dbus/dbus/dbus-uuidgen.c",
+    "./vendor/dbus/dbus/dbus-server-unix.c",
+
+    // Shared
+    "./vendor/dbus/dbus/dbus-file-unix.c",
+    "./vendor/dbus/dbus/dbus-pipe-unix.c",
+    "./vendor/dbus/dbus/dbus-sysdeps-pthread.c",
+    "./vendor/dbus/dbus/dbus-sysdeps-unix.c",
+    "./vendor/dbus/dbus/dbus-transport-unix.c",
+    "./vendor/dbus/dbus/dbus-userdb.c",
+];
+
+/// Taken directly from the libdbus meson config
+/// Reference: https://github.com/freedesktop/dbus/blob/master/meson.build#L1026
+/// Meson variable: compile_warnings
+const CWARNINGS: &'static [&'static str] = &[
+    // These warnings are intentionally disabled:
+    //  - missing field initializers being implicitly 0 is a feature,
+    //    not a bug
+    //  - -Wunused-parameter is annoying when writing callbacks that follow
+    //    a fixed signature but do not necessarily need all of its parameters
+    "-Wno-missing-field-initializers",
+    "-Wno-unused-parameter",
+    "-Wno-error=duplicated-branches",
+
+    // Under clang, need to ignore cast-align and sign-compare
+    // Reference: https://bugs.freedesktop.org/show_bug.cgi?id=10599#c5
+    "-Wno-error=cast-align",
+    "-Wno-error=sign-compare",
+
+    // Required for 1.14 but not 1.15+
+    "-Wno-error=unused-but-set-variable",
+
+    // General warnings for both C and C++
+    "-Warray-bounds",
+    "-Wchar-subscripts",
+    "-Wdouble-promotion",
+    "-Wduplicated-branches",
+    "-Wduplicated-cond",
+    "-Wfloat-equal",
+    "-Wformat-nonliteral",
+    "-Wformat-security",
+    "-Wformat=2",
+    "-Winit-self",
+    "-Winline",
+    "-Wlogical-op",
+    "-Wmissing-declarations",
+    "-Wmissing-format-attribute",
+    "-Wmissing-include-dirs",
+    "-Wmissing-noreturn",
+    "-Wnull-dereference",
+    "-Wpacked",
+    "-Wpointer-arith",
+    "-Wredundant-decls",
+    "-Wrestrict",
+    "-Wreturn-type",
+    "-Wshadow",
+    "-Wstrict-aliasing",
+    "-Wswitch-default",
+    "-Wswitch-enum",
+    "-Wundef",
+    "-Wunused-but-set-variable",
+    "-Wwrite-strings",
+
+    // Extra warnings just for C
+    "-Wdeclaration-after-statement",
+    "-Wimplicit-function-declaration",
+    "-Wjump-misses-init",
+    "-Wmissing-prototypes",
+    "-Wnested-externs",
+    "-Wold-style-definition",
+    "-Wpointer-sign",
+    "-Wstrict-prototypes",
+];
+
+/// Wrapper type for the libdbus config.
+///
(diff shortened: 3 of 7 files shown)


KillingSpark:
Without being an expert I am just wondering how the licensing works on vendored code, especially since some parts of libdbus are gpl only licenced?

Vendoring this might make it necessary to either strip all gpl Code out if possible and use the afl license or dual licence this as Apache 2 (if not using the vendoring flag) and gpl if using it.

landhb:
I'm definitely now a lawyer/expert either when it comes to license issues.

Best I could find is this answer: 

https://softwareengineering.stackexchange.com/questions/240633/is-licensing-an-issue-for-git-submodules

Which makes it sound like it's not much of an issue.

libdbus-sys/build.rs
```diff
@@ -1,8 +1,320 @@
 extern crate pkg_config;
+use std::error::Error;
+
+// Constants and gated use statements
+#[cfg(feature = "vendored")]
+mod consts {
```
diwic, line 6: There are a lot of repeated #[cfg(feature = "vendored")] in this file, how about splitting it to a separate file and include it somehow, e g through `#[path = "buildvendored.rs"] mod buildvendored;`

libdbus-sys/build.rs
```diff
@@ -1,8 +1,320 @@
...
+
+#[cfg(feature = "vendored")]
+use consts::*;
+
+/// Source Files Required
+#[cfg(feature = "vendored")]
+const SOURCE_FILES: &'static [&'static str] = &[
+    "./vendor/dbus/dbus/dbus-address.c",
+    "./vendor/dbus/dbus/dbus-auth.c",
+    "./vendor/dbus/dbus/dbus-bus.c",
```
diwic, line 31: A lot of code in this file might be copy-pasted and slightly changed from `Makefile.am` and `configure.ac`. I'm a bit worried that it makes the code a bit fragile, e g when libdbus updates and adds another source code file the vendoring will break. 

OTOH, trying to autogenerate things might be overkill...or what do you think?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libdbus-sys/vendor/dbus"]
+	path = libdbus-sys/vendor/dbus
+	url = https://github.com/freedesktop/dbus.git
```
diwic, line 3: Would it be possible to pin this to a specific tag in the repository to reduce the risk of the vendoring build breaking when the dbus.git repo changes?

libdbus-sys/build.rs
```diff
@@ -1,8 +1,320 @@
...
+        // Always sync
+        #define DBUS_USE_SYNC 1
+
+        // Unecessary
+        #undef HAVE_GETPEERUCRED
+        #undef HAVE_CMSGCRED
+    ")?;
+
+    // Define iteger types.
+    //
```
diwic, line 173: iteger -> integer

diwic (commented):
Hi @landhb and thanks for the PR! This seems like a big improvement in usability for the cross compilation use case and I really like that :-)

Apart from the review comments below, I think we should also add a Github CI/workflow for this so it doesn't break without anyone noticing.

landhb:
@diwic No problem! 100% agreed on the testing, I left the PR in draft form since I also wanted to add this to CI/CD.

I also need to make sure it builds on multiple architectures. So far I've only tested x86_64.

diwic:
@KillingSpark from https://github.com/freedesktop/dbus/blob/master/COPYING

"dbus is licensed to you under your choice of the Academic Free
License version 2.1, or the GNU General Public License version 2
(or, at your option any later version)."

From my layman's understanding, neither the AFL (which is a BSD-style license), nor the GNU GPL, make any significant differences depending on whether you link statically or dynamically.

KillingSpark:
> From my layman's understanding, neither the AFL (which is a BSD-style license), nor the GNU GPL, make any significant differences depending on whether you link statically or dynamically. 

I don't know about BSD but I think for GPL it matters for the thing about being derivative work? If it's linked statically your binary is a derivative work but if it's linked dynamically it could be combined with any abi compatible library.

But just as you I am a layman. Just wanted to raise awareness that this might be a topic to think about :)

KillingSpark:
> https://softwareengineering.stackexchange.com/questions/240633/is-licensing-an-issue-for-git-submodules

This discussion is not completely the same. This pr doesn't just use libdbus as a tool while building. I'd agree that that would be fine. It is pulling in and building the code and is necessary for using the library after it is built, making it, in my non-lawyer opinion, either a combined or derivative work.

diwic:
> I don't know about BSD but I think for GPL it matters for the thing about being derivative work?

Again in layman terms:

If you pick GPL for libdbus, and you pick MIT for dbus-rs, then the entire application will most likely be a GPL derivative. 
If you pick AFL for libdbus, and either APL or MIT for dbus-rs, then the entire application can be proprietary, open source, etc, because these licenses are not contagious.

How we link to the libdbus code does not change this.

(LGPL does make a difference between static and dynamic linking, but that does not matter here because libdbus is not licensed under LGPL, and neither is dbus-rs.)

libdbus-sys/build.rs
landhb: Good idea! Moved everything over to `build_vendored.rs` in cf25d34a9625a8b3df82b7d2686a428484a41afa.

libdbus-sys/build.rs
landhb: Yeah it was taken directly from their [Meson config](https://github.com/freedesktop/dbus/blob/master/dbus/meson.build#L31).

I think combined with pinning to a specific libdbus release (discussed in the other comment) it shouldn't break user's builds. 

So `libdbus-sys` 0.2.3 will be pinned to `libdbus` 1.15.2. And upgrading to `libdbus` 1.15.3 would require updating and testing a new `libdbus-sys` release. So broken builds should be caught during testing before we roll out another version of `libdbus-sys`.

The benefits of using the `cc` crate directly instead of calling `configure`/`cmake`/`meson` is that it doesn't swap the `libdbus-dev-1` dependency for other build-time dependencies. And allows downstream crates/users to simply use cargo to build.

.gitmodules
landhb: The submodule is actually pinned to the following git tag/release:

```
$ git submodule status
 ed866a94889e13c83dc873d8b5f86a907f908456 libdbus-sys/vendor/dbus (dbus-1.15.2)
 ```
 
 For some reason git doesn't show that in the `.gitmodules` file, only in the git object tree itself.

File: libdbus-sys/build.rs (hunk 4 above)
Comment (landhb) [reply libdbus-sys/build.rs]: Fixed https://github.com/landhb/dbus-rs/blob/vendored/libdbus-sys/build_vendored.rs#L164 thanks!

File: .gitmodules (hunk 3 above)
Comment (diwic) [reply .gitmodules:3]: I looked it up; specifying a sha or tag is not supported, so for stability I suggest we go with the 1.14 branch. You can specify `branch = 1.14` in .gitmodules. When the 1.16 branch is stable I'll be happy to take a PR for that, but the 1.15 branch is a development branch which is more likely to break from time to time, both w r t build failures and runtime bugs.

Does that make sense?

Comment (landhb) [reply .gitmodules:3]: The git-submodule is already tagged at their 1.15.2 release though. If you clone this branch of `dbus-rs` you'll be pinned at `dbus-1.15.2`.

And when we publish to crates.io it will use that tag. Similar to other sys crates, like https://github.com/alexcrichton/openssl-src-rs.

Current tag: https://github.com/freedesktop/dbus/releases/tag/dbus-1.15.2

So I don't think we need to pin to a development branch.

Comment (landhb) [reply .gitmodules:3]: Oh I see what you mean, 1.15 is a development release. Yeah we can pin to the stable 1.14 instead.

Comment (landhb) [reply .gitmodules:3]: Okay now pinned at 1.14.4. Completed in dbc134d22e78aa47cd9f7bb62c4c2e1387c2150f and 4b252a8ce0831ade5473f77b7271aa83c2400ec2.

File: libdbus-sys/build_vendored.rs [hunk 5]
```diff
@@ -0,0 +1,370 @@
...
+const DBUS_VERSION_MINOR: u32 = 15;
+const DBUS_VERSION_MICRO: u32 = 2;
+
+// Location of dbus-daemon socket
+const DBUS_SYSTEM_BUS_DEFAULT_ADDRESS: &str = "/run/dbus/system_bus_socket";
+const DBUS_SESSION_SOCKET_DIR: &str = "/tmp";
+const DBUS_MACHINE_UUID_FILE: &str = "/etc/machine-id";
+
+/// Source Files Required
+const SOURCE_FILES: &'static [&'static str] = &[
```
diwic, line 26: I was thinking, to make it as easy as possible to update this file when necessary,
Can you put references on each section? So `SOURCE_FILES` would be:
```
// From vendor/dbus/dbus/meson.build
const DBUS_LIB_SOURCES: &'static [&'static str] = &[
```

...having the same name of the constant is preferred, if this is not practical then add a comment what sections you combined. It does not have to be a step-by-step explanation or anything, just a quick one-line comment.

landhb:
@diwic The `x86_64-unknown-linux-gnu` cross build should be fixed now. That cross image has a much older gcc which didn't understand some of the `pragma` directives.

libdbus-sys/build_vendored.rs
landhb: Yeah that's a good idea. Will do.

landhb: Added in https://github.com/diwic/dbus-rs/pull/408/commits/0f1afd0b962a1d9e6043c97bff98fedd26c7933f

diwic:
@landhb Thanks, this is getting better and better :-)

Just to understand how this is all going to work for the end user, will the dbus source end up on crates.io and downloaded with the package? Or will `cross` somehow download sources from git?

libdbus-sys/build_vendored.rs
```diff
@@ -0,0 +1,389 @@
...
+    "./vendor/dbus/dbus/dbus-file-unix.c",
+    "./vendor/dbus/dbus/dbus-pipe-unix.c",
+    "./vendor/dbus/dbus/dbus-sysdeps-pthread.c",
+    "./vendor/dbus/dbus/dbus-sysdeps-unix.c",
+    "./vendor/dbus/dbus/dbus-transport-unix.c",
+    "./vendor/dbus/dbus/dbus-userdb.c",
+];
+
+/// Taken directly from the libdbus meson config
+const CWARNINGS: &'static [&'static str] = &[
```
diwic, line 96: Should also have reference

libdbus-sys/build_vendored.rs
landhb: And variable names in https://github.com/diwic/dbus-rs/pull/408/commits/94f8e697916d397b34fc44b77f5a433265c87c0b

libdbus-sys/build_vendored.rs
```diff
@@ -0,0 +1,389 @@
...
+use std::collections::hash_map::Iter;
+
+// Version of vendored libdbus
+const DBUS_VERSION: &str = "\"1.14.4\"";
+const DBUS_VERSION_MAJOR: u32 = 1;
+const DBUS_VERSION_MINOR: u32 = 14;
+const DBUS_VERSION_MICRO: u32 = 4;
+
+// Location of dbus-daemon socket
+const DBUS_SYSTEM_BUS_DEFAULT_ADDRESS: &str = "/run/dbus/system_bus_socket";
```
diwic, line 21: These three constants should also have a reference

libdbus-sys/cross_compile.md
```diff
@@ -13,6 +13,22 @@ The examples below all assume you're trying to compile for Raspberry Pi 2 or 3 r
 Cross compiling using rust-embedded/cross
 =========================================
 
+#### New Instructions:
```
diwic, line 16: Skip this line

libdbus-sys/cross_compile.md
```diff
@@ -13,6 +13,22 @@ The examples below all assume you're trying to compile for Raspberry Pi 2 or 3 r
...
+
+**Update**: With the new `vendored` feature. It is no longer required to provide a custom docker image, or have `libdbus-1-dev` installed to build. Simply enable the `vendored` feature:
```
diwic, line 18: Change to something like: 

It is not required to provide a custom docker image, or have `libdbus-1-dev` installed to build. Git will need to be installed on the build host in order to download the dbus sources. Then enable the `vendored` feature:

libdbus-sys/build_vendored.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+/// Config.h and dbus-arch-deps.h stubs (we generate config programatically
+/// instead to be passed directly to cc)
+fn generate_stubs() -> Result<(), Box<dyn Error>> {
+    let mut config = File::create("vendor/dbus/config.h")?;
+    let mut arch = File::create("vendor/dbus/dbus/dbus-arch-deps.h")?;
+
+    // Stub out the file.
+    config.write_all(b"
```
diwic, line 202: Needs reference (if there is one?)

libdbus-sys/build_vendored.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    // Define integer types.
+    //
+    // The libdbus builder relies on detecting int/long/long long width.
+    // but we can just use modern C types like uint64_t.
+    //
+    // Pointer width will still be detected at build time by our
+    // generate_config() method.
+    arch.write_all(b" 
+        #ifndef DBUS_ARCH_DEPS_H
```
diwic, line 223: Needs reference (if there is one?)

libdbus-sys/build_vendored.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        DBUS_END_DECLS
+
+        #endif /* DBUS_ARCH_DEPS_H */
+    ")?;
+    Ok(())
+}
+
+/// Generate configuration based on target platform + architecture
+fn generate_config(cc: &mut cc::Build, config: &mut Config) -> Result<(), Box<dyn Error>> {
```
diwic, line 255: Reference would be nice here too

libdbus-sys/build_vendored.rs
```diff
@@ -0,0 +1,389 @@
...
+/// enough to port to cc directly. Allowing downstream crates to rely on the
+/// dbus crate without additional host dependencies.
+pub fn build_libdbus() -> Result<(), Box<dyn Error>> {
+    let mut compiler = cc::Build::new();
+    let mut config = Config::new();
+
+    // Ensure submodule is checked out
+    if !Path::new("vendor/dbus").exists() {
+        let _ = Command::new("git")
+            .args(&["submodule", "update", "--init", "vendor/dbus"])
```
diwic, line 346: Hmm. Does this not mean that git the binary is now a build dependency of `dbus-rs`, if the vendored feature is enabled?

diwic: If installing git is a step required for cross build then it should be documented.

diwic (changes requested):
Okay, I hope this is the final set of review comments. After fixing, please squash the commits into one and I believe it should be ready for merging.

libdbus-sys/build_vendored.rs
landhb: @diwic When published on `crates.io` it won't be required.  This is more of a development convenience for people who have cloned the repository without `--recursive`.

Using `libgit2-sys` as an example.

They have the C source as a git submodule here: https://github.com/rust-lang/git2-rs/tree/master/libgit2-sys
Then when published on crates.io it's archived with the full source at the specific commit: https://docs.rs/crate/libgit2-sys/0.14.0+1.5.0/source/libgit2/

The build.rs line is something they also have, but isn't explicitly required. When cargo pulls down the `libdbus-sys` source it will include the full `dbus` source as well. So the `git` line won't run since `vendor/dbus` exists.

Their build.rs: https://github.com/rust-lang/git2-rs/blob/master/libgit2-sys/build.rs#L27

landhb:
> @landhb Thanks, this is getting better and better :-)
> 
> Just to understand how this is all going to work for the end user, will the dbus source end up on crates.io and downloaded with the package? Or will `cross` somehow download sources from git?
> 
> Edit: Found the answer myself - the build system calls `git`.

@diwic So my understanding is that publishing on `crates.io` archives the entire git submodule along with the source. Elaborated more [here](https://github.com/diwic/dbus-rs/pull/408#discussion_r1057796424) with `libgit2-sys` as an example.

libdbus-sys/cross_compile.md
landhb: Removed in 728539ee9c1a88ccf61a2c2336b168b2f99f429f

libdbus-sys/build_vendored.rs
landhb: Documented in 84b3ddf35a391205ce7340325c5b01879ced1359

libdbus-sys/build_vendored.rs
landhb: Currently no reference. The actual config is created dynamically in `generate_config()` but `config.h` is required to exist since some `dbus` source files include it.

libdbus-sys/build_vendored.rs
landhb: Also no static reference, when building with `cmake/meson` this is created. And really all it does is setup the types for `dbus_int16_t`, `dbus_int32_t`, and `dbus_int64_t`. Plus their unsigned counterparts.

libdbus-sys/build_vendored.rs
landhb: Documented in 84b3ddf35a391205ce7340325c5b01879ced1359

diwic:
@landhb 
So I merged it, then just ran (locally here, from the workspace directory) 

```
$ git pull --rebase origin
$ cargo test --features vendored
error: failed to run custom build command for `libdbus-sys v0.2.2 (/home/david/rust/dbus-rs/libdbus-sys)`

Caused by:
  process didn't exit successfully: `/home/david/rust/dbus-rs/target/debug/build/libdbus-sys-55ef2401f6ef29e5/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=build_vendored.rs

  --- stderr
  Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

Inside `libdbus-sys/vendor` there is a single `/dbus/config.h` which is 0 bytes. Any ideas? Or is this only supposed to work with cross, not normal testing/compilation?

landhb:
@diwic Yeah looks like I typo'd the source path check. Submitted a fix here: https://github.com/diwic/dbus-rs/pull/409

landhb:
Thanks for all the review work @diwic :tada:

diwic:
@landhb Thanks for the contribution! Nice to see some improvement in this area :-)

I'll wait a few days to see if any bugs come up as a result of this PR, otherwise I'll prepare a release of libdbus-sys and dbus-rs. Sounds like a plan?

diwic:
@landhb I notice tons of the following warnings in the CI output, anything to worry about?

```

libdbus-sys-linux (stable, aarch64-unknown-linux-musl)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

landhb:
Looks like it's the toolchain action: https://github.com/actions-rs/toolchain/issues/221

Since we aren't explicitly using set-output. 

Looks like [dtonly ](https://github.com/dtolnay) has a drop in replacement now, that we can swap for:

https://github.com/marketplace/actions/rustup-toolchain-install

landhb:
@diwic Started a pull here to upgrade some of those deps: https://github.com/diwic/dbus-rs/pull/411

diwic:
@landhb Released now, enjoy!

PR #410, opened 2022-12-30
2c std ownedfd

diwic:
Remove our own OwnedFd since std has one now. Backwards incompatible change, and bumps rust version requirement to 1.63.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,11 @@ Features
 
 The `futures` feature makes `dbus` depend on the `futures` crate. This enables the `nonblock` module (used by the `dbus-tokio` crate).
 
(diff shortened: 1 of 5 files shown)


diwic:
@ids1024 Hi, so was considering taking the plunge and replace our OwnedFd with std's. Here's what it would look like. Not sure why the (normal) CI fails though, do you have any idea? The rustc version should be new enough: https://github.com/actions/runner-images/blob/ubuntu22/20221212.1/images/linux/Ubuntu2204-Readme.md

ids1024:
You should import it from `std::os::unix::io` rather than `std::os::fd`. It's in `std::os::fd` now since it's available on  WASI as well as Unix, but that change was done in a later Rust release than the one that added `OwnedFd`.

diwic:
Replaced by #418

PR #411, opened 2022-12-30
Upgrade Github Actions

landhb:
Removes use of deprecated `set-output` (hopefully).


landhb:
Looks like `XAMPPRocky/get-github-release@v1` is also using `set-output`

PR #413, opened 2023-01-03
crossroads: Extend testing to cover async multi-interface GetAll

zevweiss:
This test currently fails.

### DIFF
--- a/dbus-crossroads/src/test.rs
+++ b/dbus-crossroads/src/test.rs
@@ -376,3 +376,49 @@ async fn properties_get_all_async() {
     assert_eq!(response.get("Async").unwrap().as_i64(), Some(2));
     assert_eq!(response.len(), 2);
 }
+
+#[tokio::test]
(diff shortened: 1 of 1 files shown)


zevweiss:
See https://github.com/diwic/dbus-rs/issues/414.

diwic:
Indirectly merged into master (after fixing it up - remember that tests run in parallel so bus names cannot be reused). Thanks for submitting your bug as a test case!

zevweiss:
Ah, I hadn't been aware the tests were parallelized, thanks.

However, diffing between commits ea1d92b (this PR) and a346ee378363 (the one that was merged) I don't see any differences in the test added here -- might an intended change have been omitted by mistake?

diwic:
@zevweiss it was fixed up in https://github.com/diwic/dbus-rs/commit/007b0404f6a5bc298e9561c84ad39ab23d6f35b9

Issue #412, opened 2022-12-31
Crossroads::get_children() returns not children but descendants


linuxmetel:
`Crossroads::get_children()` in [crossroads.rs](https://github.com/diwic/dbus-rs/blob/master/dbus-crossroads/src/crossroads.rs) returns not vector of all of the children, but vector of all of the descendants. For example, it returns not `vec!["com"]` but `vec!["com", "com/example"]`. Its name and its behavior do not match.

In addition, when D-Bus method `org.freedesktop.DBus.Introspectable.Introspect` is called, it returns the descendants, because of the problem. I think that it should return only the children.

Example: com.examle.dbustest has four objects: `/`, `/com`, `/com/example`, `/com/example/dbustest`.

```
$ dbus-send --print-reply=literal --session --dest=com.example.dbustest / org.freedesktop.DBus.Introspectable.Introspect | xmllint --format -
```

```xml
<?xml version="1.0"?>
<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
<node name="/">
  <interface name="org.freedesktop.DBus.Introspectable">
    <method name="Introspect">
      <arg name="xml_data" type="s" direction="out"/>
    </method>
  </interface>
  <node name="com"/>
  <node name="com/example"/>
  <node name="com/example/dbustest"/>
</node>
```

linuxmetel:
@diwic
This does not work well in some cases.

If the test is changed like following, it's not passed.

```diff
--- a/dbus-crossroads/src/test.rs
+++ b/dbus-crossroads/src/test.rs
@@ -143,6 +143,7 @@ const INTROSPECT: &str = r###"<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS O
   </interface>
   <node name="another_child_of_sample_object"/>
   <node name="child_of_sample_object"/>
+  <node name="child_of_sample_object2"/>
 </node>"###;

 #[test]
@@ -164,7 +165,9 @@ fn introspect() {
     });
     cr.insert("/com/example/sample_object0", &[token], ());
     cr.insert("/com/example/sample_object0/child_of_sample_object", &[], ());
+    cr.insert("/com/example/sample_object0/child_of_sample_object2", &[], ());
     cr.insert("/com/example/sample_object0/child_of_sample_object/subchild", &[], ());
[...]

PR #417, opened 2023-01-04
crossroads: Make sure get_children() with direct_only flag works correctly

linuxmetel:
Fixes: #412

### DIFF
--- a/dbus-crossroads/src/crossroads.rs
+++ b/dbus-crossroads/src/crossroads.rs
@@ -235,7 +235,7 @@ impl Crossroads {
             let csub: &str = &c[substart..];
             if csub.len() == 0 || csub.as_bytes()[0] != b'/' { continue; }
             let csub1 = &csub[1..];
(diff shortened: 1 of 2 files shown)


diwic:
Hi and thanks for the PR! It won't work though because it is common practice to have, say objects at `/a/b/c` without also having objects at `/a` and `/a/b`. So `a/b/c` must show up when introspecting `/`.

diwic:
Fixed in https://github.com/diwic/dbus-rs/commit/d913c608c692af6d8cdbad397cae38727c5ff4f8

linuxmetel:
@diwic
Oh, I see.

But when an object has more than one child, `prev` is not always the parent of `c` . So it's probably required to trace the paths which could be ancestors (ex. `/` , `/a` and `/a/b` for `/a/b/c` ) and check if it exists.

linuxmetel:
I'm sorry. I sent mention because of my misunderstanding. (I deleted the message)
I think that it correctly works. Thank you.

PR #424, opened 2023-01-31
DBusServer support

chayleaf:
Not tested as of now. To continue, most likely `set_timeout_functions` support will have to be added.

### DIFF
--- a/dbus/src/channel/ffichannel.rs
+++ b/dbus/src/channel/ffichannel.rs
@@ -1,106 +1,36 @@
 //! Contains some helper structs and traits common to all Connection types.-
 
-use crate::{Error, Message, to_c_str};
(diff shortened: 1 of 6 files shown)


diwic:
Hi and thanks for the PR!

I'm happy to review and merge it when it's finished (or good enough for merging). 

I looked it through quickly and something to remember w r t the callbacks from C is that they cannot panic - that will abort the process. So as a general rule of thumb, 

1) try to avoid the C callbacks completely

2) if that is not possible, review the code carefully to make sure it cannot panic, i e, do not call code we have no control over (this includes destructors)

3) if that is not possible either, surround the possibly panicking code with `catch_unwind` and handle in the best way possible

diwic:
Closing due to inactivity

PR #427, opened 2023-03-27
Provide set_sender for message

koalo:
Similar to set_serial, set_sender is useful when generating a message that was not received from a real D-Bus instance, especially for testing message processors that rely on a correctly set sender.


diwic:
Thanks!

koalo:
That was quick, thanks!

PR #441, opened 2023-06-18
Allow hiding the Introspectable interface

rhbvkleef:
## Motivation

Many real-world applications hide the existence of the `org.freedesktop.DBus.Introspectable` interface on several nodes, yet provide introspection capabilities on these nodes. This primarily happens on intermediate nodes which do not have any other functionality.

The reason for this seems to be to make DBus GUI tools nicer to use. D-Feet, or example, hides nodes that don't have any interfaces visible. That way, if you, for example, just have an object at `/com/example/test_object`, just that path will be shown, instead of three paths: `/com`, `/com/example`, and `/com/example/test_object`.

## Change Description

I added a `hidden` field to `IfaceDesc`, `MethodDesc`, `SignalDesc`, and `PropDesc`. The goal of this PR would only require it to be added to `IfaceDesc`, but for completeness sake, That property defaults to `false`, and if it is set (by the `hidden` function on the `IfaceBuilder`, `MethodDesc`, `SignalBuilder`, or `PropBuilder`.

## Why Draft?

I don't think the way I did it is the best way. Maybe instead of setting a property on the interface to hide, we should add an additional list to `Object` to represent interfaces hidden from introspection. A prior version of this PR used annotations to store the `hidden` flag, but as it is not useful to render in the introspection, that was dropped.

### DIFF
--- a/dbus-crossroads/src/crossroads.rs
+++ b/dbus-crossroads/src/crossroads.rs
@@ -13,6 +13,7 @@ use crate::utils::Dbg;
 const INTROSPECTABLE: usize = 0;
 const PROPERTIES: usize = 1;
 const OBJECT_MANAGER: usize = 2;
+const INTROSPECTABLE_HIDDEN: usize = 3;
 
 /// Contains a reference to a registered interface.
 pub struct IfaceToken<T: Send + 'static>(usize, PhantomData<&'static T>);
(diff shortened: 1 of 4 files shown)


rhbvkleef:
After also writing the proposed alternative (introducing a second field on Object (`hidden_interface`), I came to the conclusion that writing it that way is quite fragile. It is quite easy to accidentally include or exclude the `hidden_interface` field incorrectly. I still think that introducing this `hidden_interface`) field is better, as hiding the interface in the introspection is not an inherent property of an interface. I'll submit this other PR shortly in favour of this one.

PR #442, opened 2023-06-18
Allow hiding interfaces on objects

rhbvkleef:
## Motivation

Many real-world applications hide the existence of the `org.freedesktop.DBus.Introspectable` interface on several nodes, yet provide introspection capabilities on these nodes. This primarily happens on intermediate nodes which do not have any other functionality.

The reason for this seems to be to make DBus GUI tools nicer to use. D-Feet, or example, hides nodes that don't have any interfaces visible. That way, if you, for example, just have an object at `/com/example/test_object`, just that path will be shown, instead of three paths: `/com`, `/com/example`, and `/com/example/test_object`.

## Change Description

I added a `hidden_interfaces` field which is excluded for the `object_manager` and `introspectable` interfaces. All other uses use a newly introduced `all_interfaces` method on `Object`.

### DIFF
--- a/dbus-crossroads/src/crossroads.rs
+++ b/dbus-crossroads/src/crossroads.rs
@@ -8,6 +8,7 @@ use crate::ifacedesc::Registry;
 use std::collections::{BTreeMap, HashSet};
 use std::any::Any;
 use std::fmt;
+use std::iter::FromIterator;
 use crate::utils::Dbg;
 
 const INTROSPECTABLE: usize = 0;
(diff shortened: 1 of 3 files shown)


diwic:
Hi and thanks for the PR!

Before looking into it; can you tell me again if you're looking for 

 1) Hiding a path
 2) Hiding a specific interface on all paths (while showing other interfaces on those paths)
 3) Hiding a specific interface on some paths but not all of them 

Because 1) is already possible? For "intermediate" paths, just don't create them at all, and second, you can use  [set_add_standard_ifaces](https://docs.rs/dbus-crossroads/latest/dbus_crossroads/struct.Crossroads.html#method.set_add_standard_ifaces) to false if you want to control addition of the Introspectable interface yourself, it's done automatically by default out of convenience.

rhbvkleef:
I'm not looking for hiding a path, as the introspection output is useful  for these tools. Specifically, discovery of sub-nodes is important. I do want to show the introspectable interface on paths with "real" objects. So it is option 3 I'm looking for.

diwic:
@rhbvkleef Ok, but the motivation section seem to be about option 1, i e, hiding `/com` and `/com/example` entirely. 

I'm not following what the motivation would be for having option 3?

rhbvkleef:
We need to be able to obtain an introspection of /, /com, and /com/example in order to have tools be able to discover all nodes. If the interface weren't implemented at all on these paths, we wouldn't be able to locate any objects.

diwic:
> We need to be able to obtain an introspection of /, /com, and /com/example in order to have tools be able to discover all nodes. If the interface weren't implemented at all on these paths, we wouldn't be able to locate any objects.

Just add `/com/example/test_object`, and the root path will have a `<node name="com/example/test_object"/>` line, which D-feet handles that just fine. There's no need to add `/com` and `/com/example` at all.

Issue #438, opened 2023-05-30
dbus-codegen: Add attribute to specify rust type for method arguments


Snarpix:
Is it ok, if I make a PR that adds some custom attribute, to allow specifying types for generated traits, so the types can be automatically serialized\deserialized using [dbus](https://docs.rs/dbus/latest/dbus/index.html)::[arg](https://docs.rs/dbus/latest/dbus/arg/index.html)::[Append](https://docs.rs/dbus/latest/dbus/arg/trait.Append.html#) and other traits?
```
<node xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
  <interface name="com.example.MyService1.InterestingInterface">
    <method name="AddContact">
      <arg name="name" direction="in" type="s">
          <annotation name="rs.dbus.ArgType" value="crate::types::MyType"/>
      </arg>
    </method>
  </interface>
</node>
```

diwic:
Yes, I believe so - could you explain in a little more detail? E g, in the example above, what would the resulting code look like for `MyType`? Or are you just changing the method parameter type, leaving the `Append` implementation to the user?

Snarpix:
Yes, I'm just changing method parameter type, and user must implement Append, Get, and Arg.
I have successfully used custom args in pure crossroads, but when I have separate xml definition of interface, I can't do it.

diwic:
Okay, sounds good to me, go ahead and make a PR.

diwic:
Closing due to inactivity

PR #445, opened 2023-07-12
Add annotation to specify user types in generated code

Snarpix:
PR for https://github.com/diwic/dbus-rs/issues/438


diwic:
Hi! The patch looks good. Do you mind adding a test case to reduce the risk of the feature breaking with future code changes?

diwic:
Thanks! A lot more code :-) but it also helps to understand how you're supposed to use the new feature.

PR #446, opened 2023-08-07
Exclude some subdirectories from the package crate

mulkieran:


mulkieran:
Tested by:

1. Running ```cargo package --all-features```. 
2. Verifying directories excluded not in ```target/package/<packagename>```
3. ```cd ../target/package/<packagename>```.
4. Run ```cargo build --features=vendored```.
5. ```echo $?``` yields 0.

mulkieran:
@diwic I think this ought to be an extremely safe change. It excludes only files that are part of the dbus source tree pulled in by the git submodule. And exclude won't fail if directories that are excluded have stopped existing so it should work pretty nicely even if the dbus source tree changes.

diwic:
Thanks!

PR #463, opened 2024-04-04
Add a comment what return value is. 

dancerj:
I need to read through code to know what the return value for process() is.


diwic:
Thanks!

PR #465, opened 2024-05-27
Make clap dependency optional, but default & required for building the binary

AldanTanneo:
This is meant to make the `dbus-codegen` crate easier to use as a library (for instance, in build scripts) without having to pull in the clap crate.

This is not a breaking change for library users.


diwic:
Thanks!

PR #466, opened 2024-05-27
update clap version to ^4.5

AldanTanneo:


diwic:
Hi, 

I don't see why not, but it needs to be rebased on top of your other patch, which I just merged. 

Btw, what is the main reason for this change/upgrade?

AldanTanneo:
The current `clap` version that is used is very old and unmaintained, so keeping it up to date seemed wise

Also, some of the dependencies used by clap v2 are currently unmaintained or have security advisories

diwic:
> The current `clap` version that is used is very old and unmaintained, so keeping it up to date seemed wise
> 
> Also, some of the dependencies used by clap v2 are currently unmaintained or have security advisories

Okay. Rebase it and I'll merge it. Thanks!

PR #479, opened 2024-08-25
implement PartialEq<str> for string new types

ModProg:


ModProg:
not sure about allowing both w/ and w/o the trailing `\0`. As the trailing `\0` is an implementation detail, I could also remove the support for comparing with strings containing the `\0`.

diwic:
The strings are guaranteed to have a trailing nul byte (at least right now), meaning there is no need to do any special handling based on that they *maybe* have a trailing nul byte.

ModProg:
> The strings are guaranteed to have a trailing nul byte (at least right now), meaning there is no need to do any special handling based on that they _maybe_ have a trailing nul byte.

Yes, the internal string of, e.g., `Signature` has always a `\0`.

What I meant was the RHS `&str` having a trailing `\0` or not, i.e., are both `Signature::new("s") == "s" && Signature::new("s") == "s\0"` or just `Signature::new("s") == "s" && Signature::new("s") != "s\0"`.

diwic:
Right. So then I'll say that, since Signature and friends deref to str without the trailing zero, then what you should be equal to is a str without a trailing zero and nothing else.

ModProg:
Implementation now is:

    assert_eq!(Signature::new("s").unwrap(), "s");
    assert!(Signature::new("s").unwrap() != "s\0");

diwic:
Thanks!

PR #482, opened 2024-09-08
implement Append for Cow<'a, str> and Cow<'a, [T]>

### LINKED ISSUE #481: Implement `Append` for `Cow<'a, str>`
Both `&'a str` and `String` implement `Append`, but `Cow<'a, str>` doesn't implement it. Could it be added? It would enable to me avoid an allocation for my use case.

Flying-Toast:
closes #481


diwic:
Thanks! Maybe it would make sense to implement `Get` for `Cow` as well?

PR #484, opened 2024-10-01
Update tokio_server_cr.rs

stephin-thomas:
solve - "cannot find value `c` in this scope" error when running the example


diwic:
Thanks! Need to work on my test runner, this regression should not have happened...

PR #485, opened 2024-10-01
Update tokio_server_cr.rs

stephin-thomas:
remove duplicate variable declaration and moving the original declaration to right position. My previous change was to add a connection variable creation on top, I could see that it was indeed declared but at the wrong position so removing my previous change and adding the declaration to the right position.


diwic:
I moved it down in d6f530592115 to avoid a race condition, are you moving it back up again?

stephin-thomas:
I am not that good with rust, some things go above my head. In my first pull request I added this line because I didn't know the variables where created below then when I became aware of that, to avoid the duplicate creation I moved the original declaration up to where it should be. I also noticed that when running my previous pull request modification I was not getting the reply from the server now that is solved.

diwic:
I *think* this one's okay, IIRC, the critical thing is not requesting the name before setting up the receiving of incoming messages.

Issue #490, opened 2025-01-24
Add a forever blocking variant of Connection#process()


dancerj:
connection.process always takes a timeout value, and I need to give it a timeout. I wish the API took a "None" and passed it down to read_write().
Would it be a good idea to add an API variant that allows an Optional<Duration> ?

https://docs.rs/dbus/0.9.7/dbus/blocking/struct.Connection.html#method.process
https://docs.rs/dbus/0.9.7/dbus/channel/struct.Channel.html#method.blocking_pop_message

diwic:
I guess, since `read_write` takes a `Option<Duration>`. Feel free to submit a PR as long as it is backwards compatible.

diwic:
Closing (issue stalled)

PR #492, opened 2025-02-12
allow infinite wait for Connection#process() 

dancerj:
To not break existing API, create a new method with a long name

For https://github.com/diwic/dbus-rs/issues/490


dbus/src/channel/ffichannel.rs
```diff
@@ -284,8 +284,14 @@ impl Channel {
     /// Removes a message from the incoming queue, or waits until timeout if the queue is empty.
     ///
     pub fn blocking_pop_message(&self, timeout: Duration) -> Result<Option<Message>, Error> {
```
dancerj, line 286: I thought I could modify blocking_pop_message, but I haven't checked if this is exported as API

keiichiw: `Channel` trait's `blocking_pop_message` is a public API
https://docs.rs/dbus/latest/dbus/channel/struct.Channel.html#method.blocking_pop_message

dbus/src/channel/ffichannel.rs
```diff
@@ -284,8 +284,14 @@ impl Channel {
...
+        self.blocking_pop_message_with_optional_timeut(Some(timeout))
+    }
+
+    /// Removes a message from the incoming queue, or waits until timeout if the queue is empty. If timeout is None, it will wait forever.
+    ///
+    pub fn blocking_pop_message_with_optional_timeut(&self, timeout: Option<Duration>) -> Result<Option<Message>, Error> {
```
keiichiw, line 292: You also need to modify `nativechannel::Channel`.
Either `ffichannel::Channel` or `nativechannel::Channel` is chosen, depending on the feature flag [here](https://github.com/diwic/dbus-rs/blob/master/dbus/src/channel.rs#L8-L16).

nullpo-head, line 292: typo: `optional_timeout`?

diwic:
Closing (stalled)

PR #517, opened 2026-02-08
Add flexibility to codegen library usage

sveyret:
Multiple changes, each in a separate commit to answer to issue #516:
* add flexibility to code generator, to be able to generate code starting from a file, a string, D-Bus introspection, or pre-parsed definition, and to be able to directly output to a file or create a string,
* generate rust doc for created element if provided in the annotations,
* return the name of created items so that the caller can create more code around what is already generated.

The library should be backward compatible with previous version, even if the main entry point is now deprecated.


diwic:
Hi, sorry that I've been so busy that I lost this PR in the process. I'm still busy, is there anyone else who would like to step up and give a review/opinion on this PR, which is rather big for just coming without discussion first? (As suggested in the Readme file, larger changes should have a discussion before just going ahead and write a PR)

maciek134:
A bit hard to review, the majority of the changes happened in a file that was moved and reformatted.

Not sure making everything a reference is a good idea for readability, I don't think any performance you could gain from this is going to matter in a code generator. It's also completely negated by the new helper structs - I'm not sure why those were added instead of using the existing ones (`Arg` → `ArgumentDefinition` etc) - the changes seem contradictory.

If those stay, I don't think the accessors on them are necessary - they add verbosity and provide nothing in return.

It would also make sense to just use `quick_xml` + `serde` when the intermediate structs are used to store the whole parsed state before generating output - low level parsing doesn't make sense anymore.

In general the flexibility change seems to make sense in principle, but I'd take a step back and ask why? Using anything but XML for DBus interfaces is very unconventional. Any public APIs will need to be converted to XML anyway, converting your YAML to XML for the build step doesn't seem like enough of an inconvenience to change how the generator works.

The part about returning generated item names is imo not the way to go - the typical way to solve such a problem would be to let users annotate the generated structs so they can use derives or attribute macros on them - gives more flexibility to the user and makes the implementation simpler.

sveyret:
Thank you for your review, @maciek134. I will work on it when I will have time, hopefully in the coming days and give a new try. I have to say that if I am an experimented programmer, I am quite new in Rust, which can make me take some bad decisions…

> In general the flexibility change seems to make sense in principle, but I'd take a step back and ask why? Using anything but XML for DBus interfaces is very unconventional.

Yes, you’re right, but our YAML files do not only contain the DBUS interfaces, but also some other information. For example, it may contain a DBUS name to register for the service, or the name of objects to create on the interface. As they are written by hand, my colleagues will not accept to use a verbose format like XML. And the interfaces are only used between our applications, they are not exposed to public, so the XML definition is not needed.

diwic:
@maciek134  Thanks for helping out with the review!

Having read the review; here's what I have to say:

@sveyret 
Starting with anything but XML to generate dbus code is an edge use case. Whether you like XML or not, it's the standard for DBus introspection data.

So if you want your use case supported, think minimalist. What are the minimum changes you can make to the existing code, that makes it possible to start from something else than XML? E g, maybe you can use existing structs instead of creating a lot of new ones?

sveyret:
@maciek134, regarding:

> The part about returning generated item names is imo not the way to go - the typical way to solve such a problem would be to let users annotate the generated structs so they can use derives or attribute macros on them - gives more flexibility to the user and makes the implementation simpler.

I am using `codegen` in the `build.rs` script. In this same script, I also want to generate some code, using the items generated by `codegen`. Annotating the generated items would not help me in this case, or did I miss something?

Anyway, I still can add the ability to annotate the generated items, if you think that’s a good idea, event if I don’t use it.

diwic:
I'm closing this as work has stalled. Feel free to reopen in case you want to continue working on it.

PR #524, opened 2026-08-07
native-channel build for illumos

gemelen:
Initially, I bumped into the build error of the `mise` tool on a few crates, including dbus-rs, but it wasn't a dealbreaker for me, but rather a chance to look into the Rust usage for multiplatform development - I'd like this changeset to be accepted, but it's not blocking anything at this moment.

I tried to build dbus-rs on SmartOS (an illumos distributive, Solaris/OpenSolaris family), which revealed a few issues:
- use of `libc::__errno_location()` (Linux-specific)
- discrepancy between availability of `abstract` Unix sockets on Linux and other Unixen

This changeset tries to address all of these with the help of the development happened in the Rust stdlib for the Unix sockets support.

Please, review and comment. 

Rust isn't my PL of choice, so feel free to correct on the idiomatic aspects as well, thanks.


diwic:
Hi @gemelen 

I'm curious about this: dbus-native-channel is not finished, not working, and not built by default. It's abandoned until somebody wants to spend time to get it to work. 

Why do you want it to build? :-)

gemelen:
@diwic at the time of the issue created, the [mise](https://github.com/jdx/mise) tool build popped a few errors where `dbus-rs` was one of them.
I've tried to build it separately on a target platform, which also revealed this particularly unbuildable module.
I spent some time in order to build it and I'm open to share these changes.

I don't need it, but if it is of any use for the project, I'd be happy.

diwic:
@gemelen The PR submitted is for the `dbus-native-channel` crate. But the `dbus` crate does not depend on `dbus-native-channel`. The `dbus-native-channel` crate is not even on crates.io. 
So how can your patch fix compilation for dbus?

Anyhow, I think that the patch as written decreases code readability compared to original, so I'm probably going to close it.

